PR #3, opened 2017-10-13
Strip BOM from line when skipping header (webvtt)

ping:
There is a logic bug in skipping the webvtt header:
``line != "WEBVTT"`` should really be ``line == "WEBVTT"``

This worked previously because the header line had UTF-8 byte order mark characters in it.

This PR strips the BOM char before comparison and also fixes the header detection logic.

Edited to add:
Without this fix, processing ascii webvtt files will erroneously drop the first subtitle line.

### DIFF
--- a/webvtt.go
+++ b/webvtt.go
@@ -47,7 +47,8 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	// Skip the header
 	for scanner.Scan() {
 		line = scanner.Text()
-		if len(line) > 0 && line != "WEBVTT" {
+		line = strings.Trim(line, "\xef\xbb\xbf") // strip BOM chars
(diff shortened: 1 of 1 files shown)


webvtt.go
```diff
@@ -47,7 +47,8 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	// Skip the header
 	for scanner.Scan() {
 		line = scanner.Text()
-		if len(line) > 0 && line != "WEBVTT" {
+		line = strings.Trim(line, "\xef\xbb\xbf") // strip BOM chars
```
asticode: Could you use the `strings.TrimPrefix()` method and the `BOMHeader` constant instead?

ping:
Changed as requested.

asticode:
LGTM

PR #4, opened 2018-02-19
implement offset time

asnapper:
while parsing TTML, i noticed that you only support clock time format, and no offset format

see: https://www.w3.org/TR/ttaf1-dfxp/#timing-value-timeExpression

this is a "first draft" implementation

### DIFF
--- a/ttml.go
+++ b/ttml.go
@@ -31,11 +31,11 @@ const (
 var ttmlLanguageMapping = astimap.NewMap(ttmlLanguageEnglish, LanguageEnglish).
 	Set(ttmlLanguageFrench, LanguageFrench)
 
-// TTML regexp
-var ttmlRegexpDurationFrames = regexp.MustCompile("\\:[\\d]+$")
-
-// TTML offset time
-var ttmlRegexpOffsetTime = regexp.MustCompile("(\\d+)(\\.(\\d+))?(h|m|s|ms|f)$")
+// TTML Clock Time Frames and Offset Time
+var (
+	ttmlRegexpClockTimeFrames = regexp.MustCompile("\\:[\\d]+$")
+	ttmlRegexpOffsetTime      = regexp.MustCompile("^(\\d+)(\\.(\\d+))?(h|m|s|ms|f|t)$")
+)
 
 // TTMLIn represents an input TTML that must be unmarshaled
 // We split it from the output TTML as we can't add strict namespace without breaking retrocompatibility
@@ -212,6 +212,11 @@ func (d *TTMLInDuration) UnmarshalText(i []byte) (err error) {
 		metric := matches[4]
 		value, err := strconv.Atoi(matches[1])
 
+		if err != nil {
+			err = errors.Wrapf(err, "astisub: failed to parse value %s", matches[1])
+			return err
+		}
+
 		d.d = time.Duration(0)
 
 		var (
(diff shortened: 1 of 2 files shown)


ttml.go
```diff
@@ -33,6 +34,9 @@ var ttmlLanguageMapping = astimap.NewMap(ttmlLanguageEnglish, LanguageEnglish).
 // TTML regexp
 var ttmlRegexpDurationFrames = regexp.MustCompile("\\:[\\d]+$")
 
+// TTML offset time
+var ttmlRegexpOffsetTime = regexp.MustCompile("(\\d+)(\\.(\\d+))?(h|m|s|ms|f)$")
```
asticode: Doc is mentionning `t` as well as `h|m|s|ms|f`, could you add it?

asticode: Also I've the feeling the regexp should start with `^`, do you agree?

ttml.go
```diff
@@ -33,6 +34,9 @@ var ttmlLanguageMapping = astimap.NewMap(ttmlLanguageEnglish, LanguageEnglish).
 // TTML regexp
 var ttmlRegexpDurationFrames = regexp.MustCompile("\\:[\\d]+$")
```
asticode: Could you put the 2 regexps in the same var block so that it's more readable?

```go
var (
    ttmlRegexpDurationFrames = regexp.MustCompile("\\:[\\d]+$")
    ttmlRegexpOffsetTime = regexp.MustCompile("(\\d+)(\\.(\\d+))?(h|m|s|ms|f)$")
)
```

ttml.go
```diff
@@ -204,6 +208,51 @@ type TTMLInDuration struct {
...
+		value, err := strconv.Atoi(matches[1])
+
+		d.d = time.Duration(0)
+
+		var (
+			nsBase       int64
+			fraction     int
+			fractionBase float64
+		)
+		if len(matches[3]) > 0 {
```
asticode: Could you return an error in the `else` statement?

ttml.go
```diff
@@ -204,6 +208,51 @@ type TTMLInDuration struct {
...
+			fraction, err = strconv.Atoi(matches[3])
+			fractionBase = math.Pow10(len(matches[3]))
+		}
+
+		if err != nil {
+			return err
```
asticode: Can you wrap the error as well as add a prefix?

```go
if err != nil {
	err = errors.Wrap(err, "astisub: my message")
	return
}
```

ttml.go
asticode: Also could you rename `ttmlRegexpDurationFrames` to `ttmlRegexpClockTimeFrames`?

ttml.go
```diff
@@ -204,6 +208,51 @@ type TTMLInDuration struct {
...
+		case "s":
+			nsBase = time.Second.Nanoseconds()
+		case "ms":
+			nsBase = time.Millisecond.Nanoseconds()
+		case "f":
+			nsBase = time.Second.Nanoseconds()
+			d.frames = value % d.framerate
+			value = value / d.framerate
+			// TODO, fraction of frames
+		}
```
asticode: Can you add a TODO to handle `t` as well?

ttml.go
asnapper: Yes, start of string marker makes sense.

concerning ticks, i'm not quite sure how to implement it and if i'm not mistaken it would require for astisub to support a "tickrate" as well...... it would be a bit out of scope for this pull request

IMO a good compromise would be to throw a not implemented error when there is an offset-time in ticks

i'll update the PR soon :)

ttml.go
asnapper: this is not an error, if the fraction ( .001) part has 0 length, fraction and fractionBase will simply be 0

asticode:
Cheers

PR #5, opened 2018-02-28
only tests for stl umlaut

asnapper:
it's nice to write tests and then realize everything works already ;-)


asticode:
It's a good addition to the test base indeed.
Cheers

PR #6, opened 2018-03-06
STL: ignore TTI if EBN is Reserved user data

asnapper:
ignore items with EBN 254

see https://tech.ebu.ch/docs/tech/tech3264.pdf

cheers!


stl.go
```diff
@@ -157,6 +157,9 @@ const (
 	stlTimecodeStatusIntendedForUse    = "1"
 )
 
+// TTI Special Extension Block Number
+const extensionBlockNumberReserverdUserData = 254
```
asticode: There's a typo in the constant name, it should be `extensionBlockNumberReservedUserData`.

Also could you use the hexadecimal version of `254` instead ? (`0xfe`)

asnapper: done :)

PR #7, opened 2018-12-31
Added BOM-handling to SSA.

Des-Nerger:
When I tried to give astisub an .ass file containing the BOM-header I got the following error:
`astisub: line '<U+FEFF>[Script Info]' should contain at least one ':'`
This commit fixes the problem.


bom-aware-scanner.go
```diff
@@ -0,0 +1,35 @@
...
+	"bytes"
+	"io"
+)
+
+type BomAwareScanner struct {
+	*bufio.Scanner
+	lookingForBom bool
+}
+
+func NewBomAwareScanner(r io.Reader) *BomAwareScanner {
```
asticode: Can you guide me through why you need to create a specific scanner instead of just trimming the BOM byte on the first line? Is there any gain with this scanner?

ssa.go
```diff
@@ -186,9 +187,13 @@ func ReadFromSSA(i io.Reader) (o *Subtitles, err error) {
 
 		// Split on ":"
 		var split = strings.Split(line, ":")
-		if len(split) < 2 {
-			err = fmt.Errorf("astisub: line '%s' should contain at least one ':'", line)
-			return
+		if len(split) < 2 || split[0] == "" {
+			switch sectionName {
+			case ssaSectionNameScriptInfo, ssaSectionNameStyles: // Do nothing
```
asticode: Can you provide an example for this case?

ssa.go
```diff
@@ -594,14 +600,20 @@ func newSSAStyleFromString(content string, format map[int]string) (s *ssaStyle,
...
 			}
 		// Int
 		case ssaStyleFormatNameAlignment, ssaStyleFormatNameBorderStyle, ssaStyleFormatNameEncoding,
-			ssaStyleFormatNameMarginL, ssaStyleFormatNameMarginR, ssaStyleFormatNameMarginV,
-			ssaStyleFormatNameOutline, ssaStyleFormatNameShadow, ssaStyleFormatNameSpacing:
+			ssaStyleFormatNameMarginL, ssaStyleFormatNameMarginR, ssaStyleFormatNameMarginV:
 			// Parse int
 			var i int
 			if i, err = strconv.Atoi(item); err != nil {
+				fmt.Printf("%q\n", attr)
```
asticode: Can you remove this `Printf`?

subtitles.go
```diff
@@ -176,14 +176,14 @@ type StyleAttributes struct {
 	SSAMarginRight       *int // pixels
 	SSAMarginVertical    *int // pixels
 	SSAMarked            *bool
-	SSAOutline           *int // pixels
+	SSAOutline           *float64 // pixels
```
asticode: Those changes are fine but they break tests (`go test ./...`). You need to do 2 things:

- update `ssa_test.go` so that those attributes are `astiptr.Float` instead of `astiptr.Int`
- update `testdata/example-out.ssa` so that some values in the `Style` section are with `.000` at the end

webvtt.go
```diff
@@ -48,7 +48,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	for scanner.Scan() {
 		line = scanner.Text()
 		line = strings.TrimPrefix(line, string(BytesBOM))
-		if len(line) > 0 && line == "WEBVTT" {
+		if len(line) > 0 && strings.Fields(line)[0] == "WEBVTT" {
```
asticode: Nice, I didn't know about `strings.Fields`

asticode:
@Des-Nerger Nice changes! Some minor fixes needed though.

Let me know.

ssa.go
Des-Nerger: Yes. Sorry, I was mean to remove it, but forgot.

bom-aware-scanner.go
Des-Nerger: I thought it could be reused for other formats. For instance, WebVTT. But maybe I'm indeed over-engineering.

ssa.go
Des-Nerger: Here's the fragments I had problems with:
```
[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,MS UI Gothic,90,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,5,0,1,2,2,1,10,10,10,0
//
Style: Rubi,MS UI Gothic,50,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,2,1,10,10,10,0
//

[Events]
```
```
Dialogue: 0,0:23:40.69,0:23:45.56,Default,,0000,0000,0000,,東福寺駅近くにある会社からの人生の訓辞なのかなぁ～
;任天堂京都本社

:次回標題
Dialogue: 0,0:23:45.79,0:23:47.20,Default,,0000,0000,0000,,次回　「願望」
```
The specification says: `SSA will discard any lines it doesn't understand`.
and also I wanted to make the behaviour closer to libass.
Libass only reports about ill-formed lines if they are in [Fonts] or [Events] sections. Otherwise, it discard them like the rest.

bom-aware-scanner.go
Des-Nerger: Also, do you plan to make go-astisub work for other encodings, like UTF-16? I think a dedicated scanner could abstract away their differences, including different BOMs.

asticode: Making go-astisub compatible with UTF-16 is a nice idea indeed. But in the context of this PR, I'd rather keep things simple and only trim BOM bytes.

ssa.go
asticode: OK makes sense.

Could you add at least one line not understood by the SSA parser in the `testdata/example-in.ssa` so that this case is covered by the tests?

Des-Nerger: Hmm... this event line
`Unknown descriptor: discarded`
fails with this error:
`astisub: building new ssa event failed: astisub: content has 1 items whereas style format has 10 items`
So maybe instead of the `|| split[0] == ""` check, we should check if split[0] is an unknown descriptor.

bom-aware-scanner.go
Des-Nerger: Could you do it yourself? I can't come up with a nice implementation, don't want to duplicate stuff.

asticode: Sure I can do once the PR is merged.

You'll need to remove your new scanner though.

ssa.go
asticode: I think this check is fine to avoid lines without a `:`.

But as you mention there should be an additionnal check for the descriptor. I would place it [here](https://github.com/asticode/go-astisub/blob/master/ssa.go#L210) and would check the value of `header` based on the value of `sectionName` and discard lines that don't match the correct `header` value.

Does that make sense?

bom-aware-scanner.go
Des-Nerger: Ok, I have now removed it.

ssa.go
Des-Nerger: The specific `|| split[0] == ""` check is not to avoid lines without a `:`, it is to avoid lines with an empty descriptor (the `:次回標題` line above). Anyway, handling of unknown descriptors can be added later after the merge of this PR. go-astisub now works with subtitles I needed, and that's enough for me for now.

asticode:
I've merged the PR and added the BOM header trim.

Let me know if go-astisub still doesn't work with the subtitles you needed.

Cheers

PR #10, opened 2019-03-14
Support reading of non utf-8 ttml

saintberry:
Not sure if you would be interested in this change? I had to deal with some pesky TTML files that were created on windows and were encoded in utf-16le charset (ironically not valid TTML). This was semi annoying to resolve. The problem is explained in here https://groups.google.com/forum/#!topic/golang-nuts/tXcECEKC2rs

> The problem lies with encoding/xml's design: in order to use the 
> charset reader the xml library needs to examine the first line of text 
> from the xml file (where the encoding is specified). unfortunately 
> that first line contains invalid UTF-8 already, and libxml barfs 
> before it even figures out what the encoding should be to pass it to 
> our charset reader. 

This change should allow passing in utf-16 or utf-8 encoded files into the reader.


asticode:
@saintberry I've began thinking about this problem some time ago, could you try parsing your file with the code on [this branch](https://github.com/asticode/go-astisub/tree/test-ttml-non-utf8) instead?

It's basically using the `transform` package in a reader I've made [here](https://github.com/asticode/go-astitools/blob/master/unicode/reader.go#L23).

If that works, I'd rather use this method instead and I'm welcoming a PR to propagate the change to all types of subtitles.

Let me know.

saintberry:
Ahh I didn't notice that. Looks like you pushed the branch after I first looked 👍 

Could be a pebkac but I was unable to compile with that branch: 

`go-astisub@v0.0.0-20190309091915-4f82b31a8ac5/ttml.go:296:14: undefined: astiunicode.NewReader`

asticode:
Yeah sorry, it was a local change I didn't have the time to push, you couldn't have find it.

You need to `go get -u github.com/asticode/go-astitools/...` for it to compile.

saintberry:
Could you tag those commits on go-astitools? I'm having a hard time fetching go-astisub@test-ttml-non-utf8 with go-astitools@master.

Go mod refuses to fetch go-astitools@8503b2b and will only ever fetch v1.0.0 no matter what I do.

asticode:
Could you try adding a go.mod/go.sum file to `astisub` locally so that you can force astitools' commit?

asticode:
If you're having trouble testing, is there any way you can share your test files so that I can test on my end?

saintberry:
Sorry, I only just got around to making looking at this. Anyway I got it to compile with rewriting some of the go.mod files. Unfortunately the `test-ttml-non-utf8` branch didn't work with the source files I have.

```astisub: xml decoding failed: xml: encoding "UTF-16" declared but Decoder.CharsetReader is nil```

I think it's the same issue I outlined above. The XML reader doesn't like `<?xml version="1.0" encoding="UTF-16"?>`. If you change that to `<?xml version="1.0" encoding="UTF-8"?>` and _keep_ the UTF16LE BOM on the file it works with your branch. AFAIK this is an acknowledged limitation with the XML reader in go.

I've attached the file I'm using in the tests in my downstream app. Hopefully github keeps the correct encoding.

[ttml-utf16-to-vtt.ttml.zip](https://github.com/asticode/go-astisub/files/2986257/ttml-utf16-to-vtt.ttml.zip)

asticode:
Could you revert changing the go package name ?

saintberry:
This is a super old PR. Figured there wasn't appetite for this use-case. I can change the package name or just close this PR.

asticode:
I'll close this PR for now, I'll reopen it if somebody else has this use case. 

Thanks again! 

Cheers

PR #11, opened 2019-05-14
add support in Open() for upper case extensions

dlecorfec:
I encountered upper case filenames and it was more appropriate to fix the extension dispatcher rather than renaming files, as there is no need to enforce lower case filenames.


asticode:
Good idea, could you add the same logic for the `Write` method ?

dlecorfec:
Done.

asticode:
Cheers

PR #12, opened 2019-05-24
add colors to webvtt output

dlecorfec:
I looked at a TTML to WebVTT converted file from https://transcribefiles.net/other/pages/caption-subtitle-converter.htm, they used the font attribute, so here it is ... Not sure if it's OK to rely on TTMLColor but at least it has the proper value (for now).

### DIFF
--- a/webvtt.go
+++ b/webvtt.go
@@ -295,7 +295,19 @@ func (s Subtitles) WriteToWebVTT(o io.Writer) (err error) {
 
 		// Loop through lines
 		for _, l := range item.Lines {
-			c = append(c, []byte(l.String())...)
+			for _, li := range l.Items {
+				var color string
+				if li.InlineStyle != nil && li.InlineStyle.TTMLColor != "" {
+					color = li.InlineStyle.TTMLColor
+				}
+				if color != "" {
+					c = append(c, []byte("<font color=\""+color+"\">")...)
+				}
+				c = append(c, []byte(li.Text)...)
+				if color != "" {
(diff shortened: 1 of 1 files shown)


dlecorfec:
Trying a VTT subtitle with VLC, the colors don't appear when using <font color="#00ffff">, but do when using a CSS class with <c.cyan> ... Although I don't define that class anywhere. Should do some testing on various html5 and mobile players ...

asticode:
Looking at [the doc](https://www.w3.org/TR/webvtt1/#default-text-color), it seems colors should be passed through CSS classes: 

```
::cue(.yellow) {color:cyan}

02:00.000 --> 02:05.000
<c.yellow.bg_blue>This is yellow text on a blue background</c>
```

It also seems there are default values and I don't know if you can indicate a custom hexadecimal value.

Therefore, the correct way to convert TTML colors to WebVTT would be to create the proper CSS classes and add them to the proper line items instead.

Does that make sense?

dlecorfec:
Missed that doc, but I went this road eventually :) Will provide the proper
patch later, currently in Venice, nice city :)

phanirithvij:
@dlecorfec Just reminding.

asticode:
After discussing with @dlecorfec, if anybody feels like picking up this PR, please do.

Cheers

dlecorfec:
Oh well, it wasn't that hard starting from scratch ;) See https://github.com/asticode/go-astisub/pull/34

PR #14, opened 2019-08-25
webvtt,srt: Tolerate edge whitespace (and print line numbers on error)

mholt:
Hi! I'm by no means an expert of WebVTT format but this fixed a parsing problem for me. Feel free to reject if it's a bad change, or improve it or whatever you'd like.

PS. It'd be really helpful to output the line number when an error is encountered. Would you be open to a PR for that?

### DIFF
--- a/srt.go
+++ b/srt.go
@@ -33,10 +33,12 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 
 	// Scan
 	var line string
+	var lineNum int
 	var s = &Item{}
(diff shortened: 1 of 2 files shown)


asticode:
Sounds good. Actually could you add the same change to `.srt` parsing?

And I'm open for a PR displaying line numbers in errors.

Cheers

mholt:
@asticode I just put them both in this one, since they're small changes :P Hope this is good to go, thanks!

asticode:
Cheers

PR #16, opened 2019-11-07
changed stl default chars and rows value to comply with standard

flenoir:
Hi, 

I need to have width and height datas formated to standardized values regarding maximum number of characters per line and maximum number of rows in stl file metadatas. 

 Currently they are :

_maximumNumberOfDisplayableCharactersInAnyTextRow: 40,
maximumNumberOfDisplayableRows:                   23,_

which is way too much especially for Rows

I changed it to 

 _maximumNumberOfDisplayableCharactersInAnyTextRow: 37,
maximumNumberOfDisplayableRows:                   2,_

Ideally, it would be nice to be able to edit theses values through flags when converting subtitle.

As an information on standardized values, please go to this link `https://tech.ebu.ch/docs/tech/tech3360.pdf`

and page 17 under "1.4 conversion strategies" second paragraphe which explain 👍 

Specifically the Teletext format has a line length limit of 40 characters, which includes the control
characters to select colours and to activate background display for normal pages (‘Start box’
codes). Teletext pages marked as ‘sub-title’ or newsflash pages within a Teletext service should be
displayed as ‘boxed text’ by a receiver.16 The control code overhead required for boxed text has
resulted in a ‘common practise’ of using a maximum of 37 (or 36) characters in a subtitle row for
text.

and a number of row set to 2 seems to be the most used number of maximum lines

thanks a lot

 Best regards

Fabien


asticode:
I don't mind changing the default values to those you propose.

However I'd like to give the ability to the user to change those values if needed. If you check out [this line](https://github.com/asticode/go-astisub/blob/master/stl.go#L311) you'll see the `Subtitles` struct has a `Metadata` field which in turn has an attribute `STLPublisher`. It allows specifying a custom value for this value in `gsiBlock`. I think it would be a good idea to add both `STLMaximumNumberOfDisplayableCharactersInAnyTextRow` and `STLMaximumNumberOfDisplayableRows` attributes to `Metadata`, and below the line I've shared, update the `gsiBlock` values if `Metadata` values are different from `0`. Does that make sense?

flenoir:
Hi, 

Yes that’s exactlly what we need. It would be perfect if these values are editable.

Envoyé de mon iPhone

asticode:
Would you like to make those changes in this PR ? Or do you want me to do them ?

flenoir:
Hi, 

I dont know go enough to do this i guess. 

Can you do it please ?

asticode:
No problem. I've just made the changes.

After opening subtitles, you can update its `Medatada` (sometimes it's `nil` therefore you need to initialize it with `&Metadata{}`) and specifically `STLMaximumNumberOfDisplayableCharactersInAnyTextRow` and `STLMaximumNumberOfDisplayableRows` in your case. Beware as their value should be a pointer to an `Int` : you can use `astiptr.Int()` to easily create a pointer to an `Int`.

I'll close this PR for now, but let me know here if there's still something to be done.

Cheers

flenoir:
Great, thanks a lot

Envoyé de mon iPhone

PR #20, opened 2020-04-24
srt, ssa, webvtt: adds the ability to parse from strings as opposed to files on disk

robojandro:
Not too long ago I needed the ability to parse subtitles that I was accepting through an upload.

Since go-astisub does a wonderful job parsing and validating a variety of formats, I decided to keep relying on it the way I was already doing so to parse files I already on disk.

Being strapped for time, I decided to simply write the uploads to disk and proceed from there, but now that I've had a bit of time I decided that I should contribute back to project, granted that such a change would even be welcome.

This change centralizes the use of bufio.Scanner in a new function in subtitles to decouple the file reading from the parsing of the resulting lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,3 +89,25 @@ If **astisub** has been installed properly you can:
 - [x] .ssa/.ass
 - [x] .teletext
 - [ ] .smi
+
+# Parsing from strings
+
+If you are dealing with either `srt`, `ssa`, or `webvtt` subtitles as a slice of strings rather than from files on disk, you can parse them directly with the ParseFrom... functions:
+
+```go
+lines := string[]{"00:01:00.000 --> 00:02:00.000", "Credits"}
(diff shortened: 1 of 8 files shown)


asticode:
@robojandro I really do appreciate your PR but I'd rather keep the `Scanner` the way it is since it works pretty nicely with the input `io.Reader`. Indeed, if you're dealing with `[]string` instead of a file, you could use a custom `io.Reader` to use `astisub`:

```go
type myReader struct {
  currentLine string
  lines []string
}

func (r *myReader) Read(b []byte) (n int, err error) {
  // For every read, fill b with the next n bytes. Everytime you reach a new line, insert a \n byte
  return
}
```

robojandro:
Thanks for having a look.  I like that approach as well. I'll close the PR.

PR #23, opened 2020-06-17
Fix a bug parsing lines with only spaces

chadnickbok:
Adds a simple test file, `broken-in.vtt`, that starts with a single space character, followed by a newline. In master this file panics with an index out-of-bounds error. This is fixed by checking both that the line length is greater than zero, and that the resulting fields length is also greater than zero.

I've also added a very, very simple test case that can be used to confirm that the current behavior is broken in master.

### DIFF
--- a/webvtt.go
+++ b/webvtt.go
@@ -47,7 +47,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 		lineNum++
 		line = scanner.Text()
 		line = strings.TrimPrefix(line, string(BytesBOM))
-		if len(line) > 0 && strings.Fields(line)[0] == "WEBVTT" {
+		if len(line) > 0 && len(strings.Fields(line)) > 0 && strings.Fields(line)[0] == "WEBVTT" {
 			break
 		}
 	}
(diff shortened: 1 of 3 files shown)


webvtt.go
```diff
@@ -47,7 +47,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 		lineNum++
 		line = scanner.Text()
 		line = strings.TrimPrefix(line, string(BytesBOM))
-		if len(line) > 0 && strings.Fields(line)[0] == "WEBVTT" {
+		if len(line) > 0 && len(strings.Fields(line)) > 0 && strings.Fields(line)[0] == "WEBVTT" {
```
asticode, line 50: Could you simplify this line to 

```go
if fs := strings.Fields(line); len(fs) > 0 && fs[0] == "WEBVTT"
```

chadnickbok: Good suggestion, updated.

chadnickbok:
Thanks for the quick review - I've updated the PR with your suggestion and confirmed the test still passes.

asticode:
Cheers

PR #26, opened 2020-08-27
Implement tick duration support for TTML subtitles

rrooij:
I tested it with some examples and it should work. For instance, you can take the example Netflix subtitles:

```
<?xml version="1.0" encoding="UTF-8"?>
<tt xmlns="http://www.w3.org/ns/ttml" xmlns:tt="http://www.w3.org/ns/ttml" xmlns:ttm="http://www.w3.org/ns/ttml#metadata" xmlns:ttp="http://www.w3.org/ns/ttml#parameter" xmlns:tts="http://www.w3.org/ns/ttml#styling" ttp:tickRate="10000000" ttp:version="2" xml:lang="ja">
 <head>
  <styling>
   <initial tts:backgroundColor="transparent" tts:color="white" tts:fontSize="6.000vh"/>
   <style xml:id="style0" tts:textAlign="center"/>
   <style xml:id="style1" tts:textAlign="start"/>
   <style xml:id="style2" tts:ruby="container" tts:rubyPosition="auto"/>
   <style xml:id="style3" tts:ruby="base"/>
   <style xml:id="style4" tts:ruby="text"/>
   <style xml:id="style5" tts:ruby="text"/>
  </styling>
  <layout>
   <region xml:id="region0" tts:displayAlign="after"/>
  </layout>
  </head>
 <body xml:space="preserve">
  <div>
   <p xml:id="subtitle1" begin="18637368750t" end="18676157500t" region="region0" style="style0"><span style="style1">テソプ<span style="style2"><span style="style3">の所だ</span><span style="style4">カン食ン</span></span>食おう<br/><span style="style2"><span style="style3">江陵</span><span style="style5">カンヌン</span></span>で刺身でも食おう</span></p>
  </div>
 </body>
</tt>
```

(Found here: https://netflixtechblog.com/implementing-japanese-subtitles-on-netflix-c165fbe61989 )

### DIFF
--- a/ttml.go
+++ b/ttml.go
@@ -38,13 +38,13 @@ var (
 // TTMLIn represents an input TTML that must be unmarshaled
 // We split it from the output TTML as we can't add strict namespace without breaking retrocompatibility
 type TTMLIn struct {
-	Tickrate  int              `xml:"tickRate,attr"`
 	Framerate int              `xml:"frameRate,attr"`
 	Lang      string           `xml:"lang,attr"`
 	Metadata  TTMLInMetadata   `xml:"head>metadata"`
 	Regions   []TTMLInRegion   `xml:"head>layout>region"`
 	Styles    []TTMLInStyle    `xml:"head>styling>style"`
 	Subtitles []TTMLInSubtitle `xml:"body>div>p"`
+	Tickrate  int              `xml:"tickRate,attr"`
 	XMLName   xml.Name         `xml:"tt"`
 }
 
@@ -352,8 +352,8 @@ func ReadFromTTML(i io.Reader) (o *Subtitles, err error) {
 	// Loop through subtitles
 	for _, ts := range ttml.Subtitles {
 		// Init item
-		ts.Begin.tickrate = ttml.Tickrate
 		ts.Begin.framerate = ttml.Framerate
+		ts.Begin.tickrate = ttml.Tickrate
 		ts.End.framerate = ttml.Framerate
 		ts.End.tickrate = ttml.Tickrate
 
--- a/ttml.go
+++ b/ttml.go
@@ -2,7 +2,6 @@ package astisub
 
 import (
 	"encoding/xml"
-	"errors"
 	"fmt"
 	"io"
 	"math"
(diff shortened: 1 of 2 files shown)


ttml.go
```diff
@@ -39,6 +38,7 @@ var (
 // TTMLIn represents an input TTML that must be unmarshaled
 // We split it from the output TTML as we can't add strict namespace without breaking retrocompatibility
 type TTMLIn struct {
+	Tickrate  int              `xml:"tickRate,attr"`
```
asticode, line 41: Could you order attributes alphabetically ?

ttml.go
```diff
@@ -288,6 +289,10 @@ func (d TTMLInDuration) duration() time.Duration {
 	if d.framerate > 0 {
 		return d.d + time.Duration(float64(d.frames)/float64(d.framerate)*1e9)*time.Nanosecond
 	}
+	if d.tickrate > 0 {
```
asticode, line 292: Why not compute the value once and for all [here](https://github.com/asticode/go-astisub/blob/master/ttml.go#L266) ? Is the `tickrate` value not populated at that time ?

ttml.go
```diff
@@ -347,8 +352,11 @@ func ReadFromTTML(i io.Reader) (o *Subtitles, err error) {
 	// Loop through subtitles
 	for _, ts := range ttml.Subtitles {
 		// Init item
+		ts.Begin.tickrate = ttml.Tickrate
```
asticode, line 355: Can you move this line just below `ts.Begin.framerate = ttml.Framerate` ? (alphabetical order)

asticode (changes requested):
This is a nice PR.

Few changes needed though.

Also, could you add tests [here](https://github.com/asticode/go-astisub/blob/master/ttml_internal_test.go#L71) ?

ttml.go
rrooij: Unfortunately, it is not populated at that time.

rrooij:
I created the tests and fixed the styling issues (the alphabetical order).

Unfortunately the proper tick duration is still calculated in the duration function. But, I do think this is the only solution. Unmarshalltext only has access to the duration text itself and the tickrate is not known at that time. (I actually wonder if the framerate is accessible as well in that function).

Unmarshalltext should in my opinion only be responsible for extracting the text to a proper representation, not for calculating it correctly already.

asticode:
I do agree with you regarding `UnmarshalText`. I'll see what I can do.

FYI I've created a `v0.5.0` tag.

Cheers

PR #29, opened 2020-09-12
Add support to extract voicename

arun1587:
Hi, 
I added the support to extract the speaker name from the vtt text.
```
	1
	00:02:34.00 --> 00:02:35.00
	<v.first.local Roger Bingham> I'm the fist speaker

	2
	00:02:34.00 --> 00:02:35.00
	<v Bingham> I'm the second speaker
```
Should now be able to extract the speaker to  `VoiceName`  in `Line` struct.

thanks,
arun


webvtt.go
```diff
@@ -193,7 +194,9 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 			case webvttBlockNameStyle:
 				// TODO Do something with the style
 			case webvttBlockNameText:
-				item.Lines = append(item.Lines, Line{Items: []LineItem{{Text: line}}})
+				// Voice Tag to extract Speaker Name
+				voiceName, text := extractVoiceNameAndText(line)
+				item.Lines = append(item.Lines, Line{Items: []LineItem{{Text: text}}, VoiceName: voiceName})
```
asticode, line 199: I'd rather create a method that returns a `Line` directly, that way you can call directly `item.Lines = append(item.Lines, theFunction(line))`

webvtt.go
```diff
@@ -203,6 +206,50 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	return
 }
 
+func extractVoiceNameAndText(line string) (string, string) {
```
asticode, line 209: Can you rename the method to `parseTextWebVTT(i string) (o Line)` ?

webvtt.go
```diff
@@ -203,6 +206,50 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
...
+	s := scanner.Scanner{}
```
asticode, line 210: I find your implementation a bit complicated which will be a problem to for maintenance.

I'd rather use a regular expression if it works. Something like `(<v([\.\w]*)([\s\w]+)+>)*(.[^<]*)(</v>)*` should work. You get:

- voice name in index 3
- text in index 4

Declare the regexp globally and use the `FindStringSubmatch` method.

If you're not familiar with regexp in go, check out [this regexp](https://github.com/asticode/go-astisub/blob/master/ttml.go#L34)

asticode (changes requested):
Thanks for the PR, it needs some changes though

webvtt.go
arun1587: sure. I was inclined to use regexp as well (as in my first commit) but then it seemed a bit complex to figure out the expression to match all possibilities in the w3c spec for voice tag.
thanks for this `(<v([\.\w]*)([\s\w]+)+>)*(.[^<]*)(</v>)*` expression. I will update this PR sooner with the changes.

arun1587: `<v Bob>Incorrect tag?</vi>` the regexp seems to miss this case. It returns Bob. I expected to return empty string since the end tag is not a valid one`</vi>`.

webvtt.go
arun1587: done.

webvtt.go
arun1587: done

webvtt.go
asticode: Is there a specific reason why you want to catch incorrect finishing tags ?

webvtt.go
```diff
@@ -24,6 +25,7 @@ const (
 // Vars
 var (
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
+	voiceTagRegexp                     = regexp.MustCompile(`(<v([\.\w]*)([\s\w]+)+>)*(.[^<]*)(</v.*>)*`)
```
asticode, line 28: Could you rename this variable to `webVTTRegexpVoiceName` ?

webvtt.go
```diff
@@ -203,6 +205,21 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	return
 }
 
+func parseTextWebVTT(i string) (o Line) {
+	match := voiceTagRegexp.FindStringSubmatch(i)
```
asticode, line 209: you need to check that match is `!= nil` because if it's `nil` it means there are no matches

webvtt.go
```diff
@@ -203,6 +205,21 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
...
+
+	endTag := match[5]
```
asticode, line 211: You need to check that `len(match)>5`

webvtt.go
asticode, line 209: Can you rename `match` to `matches` ?

webvtt.go
```diff
@@ -203,6 +205,21 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
...
+	// should be either no tag, or match voice tag
+	if endTag != "" && endTag != "</v>" {
```
asticode, line 213: Is there a reason why you want to check the end tag and not return the text by default if the end tag is incorrect ?

asticode (changes requested):

webvtt.go
arun1587: yes

webvtt.go
arun1587: with the current regexp , the match length is always 6. so it is not really useful to check the length here.

webvtt.go
arun1587: yes

webvtt.go
arun1587: if there is incorrect closing tag, is it not a invalid content? like the html tags.

asticode: I gave this some thoughts and I think that using an `html.Tokenizer` from `golang.org/x/net/html` makes more sense for the future, since we'll surely want to capture more info in tags.

If you take this sample code

```go
tr := html.NewTokenizer(bytes.NewReader([]byte("<v.first>Hello there</v>")))
for {
	t := tr.Next()
	if err := tr.Err(); err != nil {
		log.Fatal(err)
	}
	log.Printf("type: %s - token: %s - text: %s", t, tr.Token(), tr.Text())
}
```

It will print

```
2020/09/19 10:52:10 type: StartTag - token: <v.first> - text: 
2020/09/19 10:52:10 type: Text - token: Hello there - text: 
2020/09/19 10:52:10 type: EndTag - token: </v> - text: 
2020/09/19 10:52:10 EOF
```

I think you can remove the regexp, and instead parse `html-like` tags using the same idea. It'll allow you to get the voice name.

asticode (changes requested):

webvtt.go
arun1587: The problem I see using the html tokenizer is that,
 the voice name which is considered as tag attribute is always lower-cased.
for example:
`key, attr, more := tr.TagName()` to get the VoiceName  in `"<v.first VoiceName>Hello there</v>"` would yield `voicename` instead of `VoiceName`

asticode: Using `tr.TagName()` and `tr.TagAttr()` indeed lower-case the output, but `tr.Raw()` doesn't. I would use the latter.

arun1587: yep, sounds good. please check my last commit

webvtt.go
```diff
@@ -203,6 +207,48 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	return
 }
 
+// ParseTextWebVTT parses the input line to fill the Line
+func ParseTextWebVTT(i string) (o Line) {
```
asticode, line 211: Don't export this method. 

If you're need is to be able to test it, create a `webvtt_internal_test.go' file with `package astisub` at the top. That way you'll be able to test unexported methods

asticode (changes requested):
Looks almost perfect. Just a minor change though

webvtt.go
arun1587: sounds good, done!

asticode:
Thanks for sticking with this PR, cheers

arun1587:
Sure, thanks for the nice package!

asticode:
FYI after playing around with your addition, I've updated a bit the logic (mostly it felt wrong ignoring the whole text when end tag was incorrect). Nothing fancy.

I've created a `v0.8.0` tag.

PR #34, opened 2021-01-04
webvtt: add colored output

dlecorfec:
Does the job when converting STL to WebVTT.

I think the webVTTBytes() code could probably be further refactored, to ensure it would support both font style and color, for example, but I wanted to keep this PR minimal.
At first, it could set a local variable for color, booleans for italics and other style, depending on the various settings in InlineStyle (eg. italics could come from WebVTTItalics, but also from TTMLFontStyle). Then for every variable set, print the starting vtt code, then print the text, then for every variable set, print the ending vtt code.

BTW, I didn't find a place where WebVTTItalics is set ...


asticode:
Cheers

PR #37, opened 2021-01-05
Print out the wrong number of lines when an error occurs

Scaveng:
Today, my colleague and I encountered a panic and found that an error occurred while translating the file, and the error was located in srt.go. After investigation, there was an error in a certain line of the file, but it was not printed in srt.go.For example,our srt file format has a timeline and corresponding subtitles, but if there are multiple lines of subtitles, it will panic. Under normal circumstances, it should be an error and return an error and then print the log, not panic.Therefore, we hope to add a line of code to print the number of wrong lines in srt.go to facilitate the location of errors.

### DIFF
--- a/srt.go
+++ b/srt.go
@@ -46,6 +46,12 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 
 		// Line contains time boundaries
 		if strings.Contains(line, srtTimeBoundariesSeparator) {
+			// Return the wrong number of rows
+			if len(s.Lines) == 0 {
+				err = fmt.Errorf("astisub: line %d: no lines", lineNum)
+				return
+			}
+
 			// Remove last item of previous subtitle since it's the index
 			index := s.Lines[len(s.Lines)-1]
 			s.Lines = s.Lines[:len(s.Lines)-1]


srt.go
```diff
@@ -46,6 +46,10 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 
 		// Line contains time boundaries
 		if strings.Contains(line, srtTimeBoundariesSeparator) {
+			if len(s.Lines) == 0 {
```
asticode, line 49: Could you add a comment above `// No lines`

srt.go
```diff
@@ -46,6 +46,10 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
...
+				// Print out the wrong number of lines
+				fmt.Errorf("error in the (%v) line", lineNum)
+			}
```
asticode, line 52: Could you add a line break below?

srt.go
```diff
@@ -46,6 +46,10 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 
 		// Line contains time boundaries
 		if strings.Contains(line, srtTimeBoundariesSeparator) {
+			if len(s.Lines) == 0 {
+				// Print out the wrong number of lines
```
asticode, line 50: Could you remove this comment?

srt.go
```diff
@@ -46,6 +46,10 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
...
+				fmt.Errorf("error in the (%v) line", lineNum)
```
asticode, line 51: As far as I understand it, this line won't print anything. What you want instead is:

```go
err = fmt.Errorf("astisub: line %d: no lines", lineNum)
return
```

Otherwise code will still panic and you won't have any details of the error.

asticode (changes requested):
This makes sense. Minor changes needed though

Scaveng:
ok

srt.go
```diff
@@ -46,6 +46,12 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 
 		// Line contains time boundaries
 		if strings.Contains(line, srtTimeBoundariesSeparator) {
+			// Return the wrong number of rows
+			if len(s.Lines) == 0 {
+				err = fmt.Errorf("error in the (%v) line", lineNum)
```
asticode, line 51: Can you replace this line by `err = fmt.Errorf("astisub: line %d: no lines", lineNum)`?

Indeed we want to prefix errors as well as provide more information in the message.

asticode (changes requested):

Scaveng:
ok

asticode:
Cheers

PR #39, opened 2021-02-01
stl: have vertical position no lower than 1

dlecorfec:
According to EBU 3264 (https://tech.ebu.ch/docs/tech/tech3264.pdf):
```
For teletext subtitles, VP contains a value in the range 1-23 decimal (01h-17h)
corresponding to theteletext row number of the first subtitle row.

For in-vision subtitles, VP contains a value in the range 0..NN decimal,
where NN is the maximumnumber of rows indicated in the MNR field in the GSI block
(Note: NN cannot be greater than 99 decimal(63h)).
This VP represents the number of row locations from the top of the screen to the first subtitle row.
```
My motivation: some STL subs I encountered had an erroneous zero vertical position.

### DIFF
--- a/stl.go
+++ b/stl.go
@@ -666,7 +666,11 @@ func newTTIBlock(i *Item, idx int) (t *ttiBlock) {
 
 func stlVerticalPositionFromStyle(sa *StyleAttributes) int {
 	if sa != nil && sa.STLPosition != nil {
-		return sa.STLPosition.VerticalPosition
+		vp := sa.STLPosition.VerticalPosition
+		if vp < 1 {
+			vp = 1
+		}
+		return vp
 	} else {
 		return 20
 	}


asticode:
Isn't `0` a valid position for `in-vision` subtitles ? 🤔

dlecorfec:
Ah, sorry, that's the risk of doing stuff like that at odd hours ;) Is there something in a STL file that non-ambiguously distinguish teletext from in-vision? In order to check the bound only when appropriate. Maybe with the character code tables as  seen in section 5 of the PDF?

asticode:
Mmmm I've the feeling there's no way to make the distinction between `teletext` and `in-vision` from within the code. It has to come from outside since the person using the lib is the only one knowing whether this is `teletext` or `in-vision`

dlecorfec:
Hi, just found out that Display Standard Code (DSC) in the GSI block is either 0x31 ("Level-1 teletext") or 0x32 ("Level-2 teletext") for closed (teletext) subtitles. (0x20 = undefined, 0x30 = open/invision).

asticode:
Great ! Do you have a link with that info by any chance?

dlecorfec:
On page 6 of https://tech.ebu.ch/docs/tech/tech3264.pdf

asticode:
Damn, constants [already exist](https://github.com/asticode/go-astisub/blob/master/stl.go#L123), I forgot about that.

Then we have our way to distinguish between open subtitles and teletext.

dlecorfec:
Better PR incoming, I close this one.

PR #43, opened 2021-06-10
TTML: Fixed improper spacing between captions

discovery-avishekgulshan:
1. Fixed improper spacing between captions while converting TTML -> VTT
2. Converted TTML style attributes to pointers
3. Fixed Webvtt out not using parent style attributes in the "region" section


subtitles.go
```diff
@@ -327,7 +339,9 @@ func (l Line) String() string {
 	for _, i := range l.Items {
 		texts = append(texts, i.Text)
 	}
-	return strings.Join(texts, " ")
+	t := strings.Join(texts, " ")
+	t = strings.ReplaceAll(t, "  ", " ")
```
asticode, line 343: If you're trying to replace multi spaces into single spaces, I think you'd be better off using a regexp

ttml.go
```diff
@@ -70,35 +70,48 @@ type TTMLInMetadata struct {
...
+}
+
+// StyleAttributes converts TTMLInStyleAttributes for region into a StyleAttributes
+func (s TTMLInStyleAttributes) regionStyleAttributes() (o *StyleAttributes) {
+	o = getStyleAttributes(s)
+	o.propagateTTMLRegionAttributes()
+	return
+}
+
+func getStyleAttributes(s TTMLInStyleAttributes) *StyleAttributes {
```
asticode, line 113: I'd rather rename this function to something more TTML-specific and more self-explanatory like `newStyleAttributesFromTTMLInStyleAttributes`

util/util.go
```diff
@@ -0,0 +1,5 @@
+package util
+
+func StringToPointer(str string) *string {
```
asticode, line 3: Can you remove this function and use [this one](https://github.com/asticode/go-astikit/blob/master/ptr.go#L41) instead (`astikit` is already used in the lib)

subtitles.go
asticode, line 343: Also, can you provide an example `ttml` that was having issue with that bit of code?

ttml.go
```diff
@@ -394,7 +405,7 @@ func ReadFromTTML(i io.Reader) (o *Subtitles, err error) {
 			// New line decoded as a line break. This can happen if there's a "br" tag within the text since
 			// since the go xml unmarshaler will unmarshal a "br" tag as a line break if the field has the
 			// chardata xml tag.
-			for idx, li := range strings.Split(tt.Text, "\n") {
+			for idx, li := range strings.Split(tt.Text, "\\n") {
```
asticode, line 408: Can you provide an example `ttml` that was having issue with that bit of code?

ttml.go
```diff
@@ -421,6 +432,15 @@ func ReadFromTTML(i io.Reader) (o *Subtitles, err error) {
 			}
 
 		}
+		//if multiple items in `l` are present without new line then adding space at end so it can become readable after concatenation
+		if len(l.Items) > 0 {
```
asticode, line 436: Can you provide an example `ttml` that was having issue with that bit of code?

asticode (commented):
I know this is only a draft, but I thought it could still be productive to share my early comments

ttml.go
discovery-avishekgulshan: done

util/util.go
discovery-avishekgulshan: done

subtitles.go
discovery-avishekgulshan: I used the same ttml in testdata i.e. example-in.ttml

While converting from TTML -> VTT 
TTML: 
` <p xml:id="sub_2" begin="00:02:04.080" end="00:02:07.120" region="region_2">
                MAN:
                <tt:br/>
                How did we <span style="style_1" color="green">end up</span> here?
            </p>`
 VTT:
Expected: How did we end up here?
Actual: How did weend uphere?

ttml.go
discovery-avishekgulshan: I used testdata>example-in.ttml, added '\n' n tested, it was not working.

ttml.go
discovery-avishekgulshan: I used testdata/example-in.ttml

After adding this fix, SRT, and STL output added this additional space while joining the texts, so I added code to replace extra space while joining the Text for SRT and STL.

subtitles.go
discovery-avishekgulshan: done

asticode:
I'm OK with `Converted TTML style attributes to pointers` and `Fixed Webvtt out not using parent style attributes in the "region" section` but I'm still wondering what was wrong with `Fixed improper spacing between captions while converting TTML -> VTT`.

Can you share one or more `ttml` that were causing issue before your changes?

asticode:
Never mind, I got it

subtitles.go
```diff
@@ -327,7 +340,9 @@ func (l Line) String() string {
 	for _, i := range l.Items {
 		texts = append(texts, i.Text)
 	}
-	return strings.Join(texts, " ")
+	t := strings.Join(texts, " ")
```
asticode, line 343: This feels hacky, I think if we implement the rule `trim spaces on input, add in-between spaces on output`, we don't need this logic anymore

ttml.go
asticode, line 436: I think if we implement the rule `trim spaces on input, add in-between spaces on output`, we don't need this logic anymore.

However you can add this logic on output [here](https://github.com/asticode/go-astisub/blob/master/ttml.go#L643) to make sure we add `lineItem.Text + ""` when `idx < len(item.Lines) -1`

ttml.go
asticode: What do you mean `it was not working`? What did you expect to see?

asticode (changes requested):
I think implementing the rule `trim spaces on input, add in-between spaces on output` is a better idea.

It would mean making sure to:

- trim space [here](https://github.com/asticode/go-astisub/blob/master/srt.go#L107) and [here](https://github.com/asticode/go-astisub/blob/master/ssa.go#L1017) (I think spaces are already trimmed on input for all other formats)
- add in-between spaces [here](https://github.com/asticode/go-astisub/blob/master/ssa.go#L930) (this is a bug here, you need to replace `""` with `" "`), [here](https://github.com/asticode/go-astisub/blob/master/ttml.go#L643) and [here](https://github.com/asticode/go-astisub/blob/master/webvtt.go#L400) (I think in-between spaces are already added for all other formats)

ttml.go
discovery-avishekgulshan: The code is supposed to spilt the string on '\n' but it was not splitting so I added this extra '\' and now it's working.

stl.go
```diff
@@ -673,7 +673,7 @@ func stlVerticalPositionFromStyle(sa *StyleAttributes) int {
 }
 
 func (li LineItem) STLString() string {
-	rs := li.Text
+	rs := strings.TrimSpace(li.Text)
```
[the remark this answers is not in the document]
discovery-avishekgulshan: Thanks for your help, did all the changes

subtitles.go
discovery-avishekgulshan: yep, did felt hacky. Removed it

ttml.go
asticode: Mmmm I'm really not sure `.ttml` text should be split on string `\n` 🤔 I think it should only be split on line returns.

Can you share some official docs or RFC that provide the same information as you?

asticode (changes requested):
You forgot to add in-between space [here](https://github.com/asticode/go-astisub/blob/master/ttml.go#L643)

ttml.go
discovery-avishekgulshan: I also think the same, but the code was written to split on new line characters, and during my testing, this code to split was not working, so I added this extra slash.
Anyway, removing it.

asticode:
Can you juste remove changes to `go.sum` from this PR?

Then I'll merge the PR.

discovery-avishekgulshan:
> Can you juste remove changes to `go.sum` from this PR?
> 
> Then I'll merge the PR.

go.sum automatically got updated when I ran go mod download. Trying to replace with the original file.

asticode:
Thanks for sticking with this PR.

FYI I've created a `v0.13.0` tag.

Cheers

Issue #49, opened 2021-07-19
Setting STLTimecodeStartOfProgramme to zero before decoding STL?


TobbeEdgeware:
Thanks for a very nice and useful library!

I've run into STL files where STLTimecodeStartOfProgramme is set to 10 hours while the actual timestamps start at 0. As far as I've understood, this is a "safety mechanism" to avoid non-processed insertion of subtitles.

Since the offset is subtracted from the timestamp at line 255 in stl.go, I end up with large negative timestamps:

https://github.com/asticode/go-astisub/blob/8c108a572cf9c538fa7fe5a86f85f72e948a681d/stl.go#L255

Looping through all cues and adding 10hours to each of them should give the correct values back, but it would be much more efficient and more transparent to have an option to set `STLTimecodeStartOfProgramme = 0` in the call to ReadFromSTL, or have this done automatically if the timestamps are less than the `STLTimecodeStartOfProgramme`. 

It's not clear to me what is the best way to progress here, but in order to not break the interface, maybe some optional parameters like in 

https://github.com/edgeware/mp4ff/blob/bb9320744777dc97f18034c8aed45a9bcdbaa995/mp4/file.go#L121

could be a viable alternative?

asticode:
I don't mind breaking the signature of `ReadFromSTL` and I would basically do the same thing as for [teletext](https://github.com/asticode/go-astisub/blob/master/teletext.go#L335), by adding a `STLOptions` struct containing an `IgnoreTimecodeStartOfProgramme` attribute

TobbeEdgeware:
Unfortunately, PR #50  was incorrect since it is not `o.Metadata.STLTimecodeStartOfProgramme` that is subtracted from the timestamps, but `g.timecodeStartOfProgramme`. I'll try make a new PR which includes a test to verify that the new option is really working. Sorry about that.

TobbeEdgeware:
@asticode Thanks for the quick merge and the new bug-fix release!

PR #50, opened 2021-07-20
Add STL option to ignore start of programme timecode

TobbeEdgeware:
Solves #49 using an option parameter in STLOptions similar to what is available for Teletext.

### DIFF
--- a/stl.go
+++ b/stl.go
@@ -213,11 +213,10 @@ func ReadFromSTL(i io.Reader, opts STLOptions) (o *Subtitles, err error) {
 		STLPublisher:                                        g.publisher,
 		STLRevisionDate:                                     &g.revisionDate,
 		STLSubtitleListReferenceCode:                        g.subtitleListReferenceCode,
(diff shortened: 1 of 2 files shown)


stl.go
```diff
@@ -210,6 +216,9 @@ func ReadFromSTL(i io.Reader) (o *Subtitles, err error) {
 		STLTimecodeStartOfProgramme:                         g.timecodeStartOfProgramme,
 		Title:                                               g.originalProgramTitle,
 	}
+	if opts.IgnoreTimecodeStartOfProgramme {
```
asticode, line 219: I'd rather remove line 216 and instead add

```go
if !opts.IgnoreTimecodeStartOfProgramme {
  o.Metadata.STLTimecodeStartOfProgramme = g.timecodeStartOfProgramme
}
```

asticode (changes requested):
Just a minor change needed

stl.go
TobbeEdgeware: OK. Makes sense. I've changed that.

asticode:
FYI I've created a `v0.15.0` tag

PR #52, opened 2021-07-22
fix: ignore STL timecodeStartOfProgram and test that it works

TobbeEdgeware:
A proper fix of #49 including a test with test content.


asticode:
LGTM

FYI I've created a `v0.15.1` tag

PR #58, opened 2021-10-17
add a debug option to astisub to output informations on a subtitle

dlecorfec:
Hello!

This PR adds a "-d" option to the astisub command, which sets a bool in the option struct passed to Open().
On reading a STL file, outputs to stdout:
* some values of the GSI block
* subtitle lines with their number, start time, end time, vertical position and number of line (and exclamation marks if something odd in timestamps)
* the total number of errors encountered

It is designed for the interactive human user, no concern has been given to automatic parsing of this output.

```
STL:GSIBlock
STL:  DisplayStandardCode:0x31
STL:  TotalNumberOfTTIBlocks:959
STL:  TotalNumberOfSubtitleGroups:1
STL:  TotalNumberOfSubtitles:959
STL:
STL: #0000   00000.080 - 00002.000   vp=22	lines=1 [*Thème musical de l'émission]
STL: #0001   00002.200 - 00004.800   vp=22	lines=1 [...]
STL: #0002   00005.000 - 00008.760   vp=20	lines=2 [-Mais oui ! Mais oui, les petits potes sont là,]
[...]
STL: #0378 ! 00990.560 - 00988.640 ! vp=16	lines=1 [avant celui pour le public belge.]
[...]
STL: #0958   02614.560 - 02618.400   vp=20	lines=2 [Sous-titrage ST' 501]
STL: 1 error(s)
```

Maybe it would be a better idea to do the lines output at the end of subtitles.go:Open, so it would be common to all kinds of subtitles and not clutter stl.go.

An unoprotected global struct keeps the debug state: debug is not really meant to be enabled when using astisub as a lib.

Not sure if it's worth being incorporated in astisub, but I keep going back to that debug output when problems are signaled in our subs ;)

### DIFF
--- a/astisub/main.go
+++ b/astisub/main.go
@@ -15,6 +15,7 @@ var (
 	teletextPage     = flag.Int("p", 0, "the teletext page")
 	outputPath       = flag.String("o", "", "the output path")
 	syncDuration     = flag.Duration("s", 0, "the sync duration")
+	debug            = flag.Bool("d", false, "show debug informations")
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


asticode:
I'll close this PR for now but keep it in mind in case we come up with a better idea. Thanks again!

PR #61, opened 2021-12-04
Use Items.Index in VTT write method

mysamimi:
currently use array index vs item.index. I think this better :)
I add method FixIndex, 4 reindex Items. this useful after call fragments & unfragments


sylwek78 (approved):

mysamimi:
resolve conflicts

PR #63, opened 2022-01-13
webvtt: Handle X-TIMESTAMP-MAP header

anupcshan:
When specified, X-TIMESTAMP-MAP should be used to compute an offset for
each subtitle record in the file.

### DIFF
--- a/webvtt.go
+++ b/webvtt.go
@@ -3,6 +3,7 @@ package astisub
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"regexp"
@@ -23,6 +24,7 @@ const (
 	webvttBlockNameStyle          = "style"
 	webvttBlockNameText           = "text"
 	webvttTimeBoundariesSeparator = " --> "
+	webvttTimestampMap            = "X-TIMESTAMP-MAP"
 )
 
 // Vars
@@ -36,6 +38,46 @@ var (
 // parseDurationWebVTT parses a .vtt duration
 func parseDurationWebVTT(i string) (time.Duration, error) {
 	return parseDuration(i, ".", 3)
+}
+
+// https://tools.ietf.org/html/rfc8216#section-3.5
+// Eg., `X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:900000` => 10s
+//      `X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:180000` => 2s
+func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error) {
+	splits := strings.Split(line, "=")
+	if len(splits) <= 1 {
+		err = fmt.Errorf("astisub: invalid X-TIMESTAMP-MAP, no '=' found")
+		return
+	}
+	right := splits[1]
+
+	var local time.Duration
+	var mpegts int64
+	for _, split := range strings.Split(right, ",") {
+		splits := strings.SplitN(split, ":", 2)
+		if len(splits) <= 1 {
+			err = fmt.Errorf("astisub: invalid X-TIMESTAMP-MAP, part %q didn't contain ':'", right)
+			return
+		}
+
+		switch strings.ToLower(strings.TrimSpace(splits[0])) {
+		case "local":
+			local, err = parseDurationWebVTT(splits[1])
+			if err != nil {
+				err = fmt.Errorf("astisub: parsing webvtt duration failed: %w", err)
+				return
+			}
+		case "mpegts":
+			mpegts, err = strconv.ParseInt(splits[1], 10, 0)
+			if err != nil {
+				err = fmt.Errorf("astisub: parsing int %s failed: %w", splits[1], err)
+				return
+			}
+		}
+	}
+
+	timeOffset = time.Duration(mpegts)*time.Second/90000 - local
+	return
 }
 
 // ReadFromWebVTT parses a .vtt content
@@ -63,6 +105,8 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	var blockName string
 	var comments []string
 	var index int
+	var timeOffset time.Duration
+
 	for scanner.Scan() {
 		// Fetch line
 		line = strings.TrimSpace(scanner.Text())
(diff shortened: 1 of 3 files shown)


webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
 	return parseDuration(i, ".", 3)
 }
 
+// https://tools.ietf.org/html/rfc8216#section-3.5
+func computeOffsetFromTimestampMap(line string) (time.Duration, error) {
```
asticode, line 43: Could you rename this function to `parseTimestampMapWebVTT`

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+	var timeOffset time.Duration
+
+	right := strings.Split(line, "=")[1]
```
asticode, line 46: You should check whether `len(strings.Split(line, "=")) > 0` otherwise it may panic here

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+	var local time.Duration
+	var mpegts int64
+	for _, split := range strings.Split(right, ",") {
+		splits := strings.SplitN(split, ":", 2)
```
asticode, line 50: You should check `len(splits) > 1` otherwise it may panic below

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+		switch splits[0] {
```
asticode, line 51: could you switch on `string.ToLowerCase(splits[0])` instead ? and replace `case` values with their lower case version ? that way the lib will handle `LOCAL` as well as `local` for instance

webvtt.go
asticode, line 43: Could you use named return arguments : `(timeOffset time.Duration, err error)`?

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
 	return parseDuration(i, ".", 3)
 }
 
+// https://tools.ietf.org/html/rfc8216#section-3.5
+func computeOffsetFromTimestampMap(line string) (time.Duration, error) {
+	var timeOffset time.Duration
```
asticode, line 44: You can remove this line if we use named return arguments

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+
+	right := strings.Split(line, "=")[1]
+	var local time.Duration
+	var mpegts int64
+	for _, split := range strings.Split(right, ",") {
+		splits := strings.SplitN(split, ":", 2)
+		switch splits[0] {
+		case "LOCAL":
+			var err error
```
asticode, line 53: You can remove this line if we use named return arguments

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+			local, err = parseDurationWebVTT(splits[1])
+			if err != nil {
+				return timeOffset, err
+			}
+		case "MPEGTS":
+			var err error
```
asticode, line 59: You can remove this line if we use named return arguments

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+	var local time.Duration
+	var mpegts int64
+	for _, split := range strings.Split(right, ",") {
+		splits := strings.SplitN(split, ":", 2)
+		switch splits[0] {
+		case "LOCAL":
+			var err error
+			local, err = parseDurationWebVTT(splits[1])
+			if err != nil {
+				return timeOffset, err
```
asticode, line 56: Could you wrap the error : 

```go
if local, err = parseDurationWebVTT(splits[1]); err != nil {
  err = fmt.Errorf("astisub: parsing webvtt duration failed: %w", err)
  return
}
```

webvtt.go
```diff
@@ -38,6 +39,35 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
...
+			}
+		case "MPEGTS":
+			var err error
+			mpegts, err = strconv.ParseInt(splits[1], 10, 0)
```
asticode, line 60: Could you wrap the error : 

```go
if mpegts, err = strconv.ParseInt(splits[1], 10, 0); err != nil {
  err = fmt.Errorf("astisub: parsing int %s failed: %w", splits[1], err)
  return
}
```

webvtt.go
```diff
@@ -190,6 +222,18 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 
 			// Append item
 			o.Items = append(o.Items, item)
+
+		case strings.HasPrefix(line, webvttTimestampMap):
+			if len(item.Lines) > 0 {
+				err = fmt.Errorf("Found timestamp map after processing subtitle items")
```
asticode, line 228: Can you use `errors.New` instead, prefix error with `astisub: ` and start with a lower case letter ? `astisub: found timestamp map after processing subtitle items`

webvtt.go
```diff
@@ -190,6 +222,18 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
...
+				return
+			}
+
+			timeOffset, err = computeOffsetFromTimestampMap(line)
```
asticode, line 232: Could you wrap the error : 

```go
if timeOffset, err = parseTimestampMapWebVTT(line); err != nil {
  err = fmt.Errorf("astisub: parsing webvtt timestamp map failed: %w", err)
  return
}
```

webvtt_internal_test.go
```diff
@@ -35,3 +37,42 @@ func TestParseTextWebVTT(t *testing.T) {
...
 	})
 }
+
+func TestTimestampMap(t *testing.T) {
+	for i, c := range []struct {
+		line           string
+		expectedOffset time.Duration
+		expectError    bool
+	}{
+		{
```
asticode, line 47: Could you add cases where :

- `local` is not empty and `mpegts` is empty
- both `local` and `mpegts` are not empty

asticode (changes requested):
This is a great PR ! ❤️ 

Some changes needed though.

Also, could add a test inspired by [this one](https://github.com/asticode/go-astisub/blob/master/webvtt_test.go#L49) to test a simple full webvtt file with only a couple of items and a timestamp map, to test it's taken into account?

anupcshan:
> This is a great PR ! heart
> 
> Some changes needed though.
> 
> Also, could add a test inspired by [this one](https://github.com/asticode/go-astisub/blob/master/webvtt_test.go#L49) to test a simple full webvtt file with only a couple of items and a timestamp map, to test it's taken into account?

Thanks for the detailed review! I've made all the suggested changes and added a test with a full webvtt.

asticode (approved):
❤️

asticode:
FYI I've added a `v0.20.0` tag

PR #70, opened 2022-11-07
support all characters in cue voice span

finn0:
Hi, go-astisub provides great help for me, thank you for the hard work!
I made a PR so it can recognize non-ASCII characters in the cue voice span, please help to double check, thanks.

### DIFF
--- a/webvtt.go
+++ b/webvtt.go
@@ -32,7 +32,7 @@ var (
 	bytesWebVTTItalicEndTag            = []byte("</i>")
 	bytesWebVTTItalicStartTag          = []byte("<i>")
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
-	webVTTRegexpStartTag               = regexp.MustCompile(`(<v([\.\w]*)([\s\w]+)+>)`)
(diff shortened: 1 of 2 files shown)


webvtt.go
```diff
@@ -32,7 +32,7 @@ var (
 	bytesWebVTTItalicEndTag            = []byte("</i>")
 	bytesWebVTTItalicStartTag          = []byte("<i>")
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
-	webVTTRegexpStartTag               = regexp.MustCompile(`(<v([\.\w]*)([\s\w]+)+>)`)
+	webVTTRegexpStartTag               = regexp.MustCompile(`(<v([\.\w]*)(.{1,}?)>)`)
```
asticode, line 35: Could you replace `{1,}` with `+` ?

asticode (changes requested):
Nice PR, one tiny change needed though 👍

webvtt.go
finn0: Thank you for the suggestion. Pushed a new commit.

asticode:
fyi I've created a `v0.23.0` tag

PR #75, opened 2023-01-20
Improve STL GSIBlock write

dlecorfec:
This PR allows astisub to write a STL file with a GSIBlock closer to the STL file read, by adding the copy of the fields revisionNumber, translatedEpisodeTitle, translatedProgramTitle, translatorContactDetails and translatorName to/from the Subtitles.Metadata. Also the translatedEpisodeTitle and translatedProgramTitle were in reverse order in the output block (I checked in the EBU spec).

Why? I'm investigating why the write to STL doesn't seem to produce a valid STL (valid tcin/tcout but no text when converting that STL back to WebVTT), so I try to get the `convert -i in.stl -o out.stl` to produce a close (perfect?) copy, so I'm starting with the GSIBlock ...


asticode:
Thanks for the PR!

Tests seem broken, you'll need to fix them as well. Let me know if you need some guidance for that 👍

dlecorfec:
Oops sorry! This should do the trick.

asticode:
Cheers! ❤️

PR #76, opened 2023-01-22
STL: preserve editor's name and contact details, and TTI block justification

dlecorfec:
Preserve the editor's name and editor's contact details when copying a GSI block, and preserve the justification when copying the TTI blocks.

The example-opn-out.stl had to be patched to have the same TTI justifications as the example-opn-in.stl.


stl.go
```diff
@@ -218,6 +218,8 @@ func ReadFromSTL(i io.Reader, opts STLOptions) (o *Subtitles, err error) {
 		STLMaximumNumberOfDisplayableRows:                   astikit.IntPtr(g.maximumNumberOfDisplayableRows),
 		STLOriginalEpisodeTitle:                             g.originalEpisodeTitle,
 		STLPublisher:                                        g.publisher,
+		STLEditorName:                                       g.editorName,
```
asticode, line 221: Could you order this alphabetically?

stl.go
```diff
@@ -390,6 +392,8 @@ func newGSIBlock(s Subtitles) (g *gsiBlock) {
 		}
 		g.originalEpisodeTitle = s.Metadata.STLOriginalEpisodeTitle
 		g.publisher = s.Metadata.STLPublisher
+		g.editorName = s.Metadata.STLEditorName
```
asticode, line 395: Could you order this alphabetically?

subtitles.go
```diff
@@ -337,6 +337,8 @@ type Metadata struct {
 	STLMaximumNumberOfDisplayableRows                   *int
 	STLOriginalEpisodeTitle                             string
 	STLPublisher                                        string
+	STLEditorName                                       string
```
asticode, line 340: Could you order this alphabetically?

stl.go
```diff
@@ -677,6 +681,19 @@ func newTTIBlock(i *Item, idx int) (t *ttiBlock) {
 		verticalPosition:     stlVerticalPositionFromStyle(i.InlineStyle),
 	}
 
+	if i.InlineStyle != nil && i.InlineStyle.STLJustification != nil {
```
asticode, line 684: Could you create a `stlJustificationCodeFromStyle` method with this logic inside instead? A bit like `stlVerticalPositionFromStyle()`

asticode (changes requested):

stl.go
```diff
@@ -707,6 +694,24 @@ func newTTIBlock(i *Item, idx int) (t *ttiBlock) {
 	return
 }
 
+func stlJustificationCodeFromStyle(sa *StyleAttributes) byte {
+	if sa == nil || sa.STLPosition == nil {
```
asticode, line 698: I think you want to check `sa.STLJustification == nil` instead of `sa.STLPosition == nil`

dlecorfec: Duh sorry 🤦

asticode:
Cheers

PR #87, opened 2023-06-20
Propagate STL vertical position to WebVTT

dlecorfec:
It should work with both teletext (Max Number of Rows = 23) and in-vision (MNR=0..99) subtitles, but only tested on teletext subtitles (no-op if MNR == 0 to avoid zero division).
It propagates the STL VP to the WebVTT line attribute, as percentage, by diving the row number by the max number of rows.
For example, if VP = 22 in a teletext TTI, then percentage = int(22*100/23) = 95, thus it will output "line:95%" in the VTT file.

### DIFF
--- a/subtitles.go
+++ b/subtitles.go
@@ -261,6 +261,9 @@ func (sa *StyleAttributes) propagateSTLAttributes() {
 			sa.WebVTTAlign = "left"
 		}
 	}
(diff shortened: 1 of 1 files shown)


asticode:
Thanks for the PR ❤️ 

Let me know if you need a tag 👍

dlecorfec:
Thanks! I'd like a little tag please :)

asticode:
`v0.25.0`

Issue #85, opened 2023-04-25
Escape problem when convert from other formats to VTT, and from VTT to other formats


NhanNguyen700:
Hi,

I tried to convert some SRT to VTT, and if the SRT file contain some special character like &, it is not escaped in output VTT:
- Test case convert SRT to VTT:
   Input, SRT sample:
   ```
   1
   00:00:03,400 --> 00:00:06,177
   This is a sentence with &
   ```

   Expected VTT:
   ```                                                                                              
   WEBVTT

   1
   00:00:03.400 --> 00:00:06.177
   This is a sentence with &amp;
   ```

   Actual VTT:
   ```                                                                                              
   WEBVTT

   1
   00:00:03.400 --> 00:00:06.177
   This is a sentence with &
   ```
The same problem happen when I convert from TTML to VTT

And when I convert from VTT to SRT, I expect that `&amp;` in VTT will be changed to `&` in SRT, but nothing change, `&amp;` is brought to SRT without unescaping.

tobbee:
A single & is not a valid character in WebVTT, so it should not be automatically escaped. One could argue that an error should be raised.

Accordiing to WebVTT spec, the following should happen if a & is encountered:

_U+0026 AMPERSAND (&)
Set tokenizer state to the [HTML character reference in data state](https://www.w3.org/TR/webvtt1/#html-character-reference-in-data-state), and jump to the step labeled next._

SRT is much less well-specified, so it is  not clear what should happen, but I think it is reasonable to do the same thing as for WebVTT, and not automatically change from `&` to `&amp;` or vice versa.

Therefore I don't think this should be reported as a bug and be dropped.

asticode:
I agree with @tobbee 👍

NhanNguyen700:
I don't get it, that document is about reading WebVTT from a WebVTT file, right? As I understand, if there is a text like this: `I &amp; you are talking` in a WebVTT file, when a program read it and print the text to console, the console should display: `I & you are talking`, right? `&amp;` is an character reference in HTML spec https://html.spec.whatwg.org/multipage/syntax.html#character-references, and it should be unescaped to `&` when being displayed. But as long as I see, It prints `I &amp; you are talking` to the console.

[...]

PR #89, opened 2023-07-20
escape-unescape '<' character for WebVTT

NhanNguyen700:
Hi all,

We have already escape "&" because of my reporting in this ticket https://github.com/asticode/go-astisub/issues/85, but we are lacking of escape-unescape for "<", with escape-unescape "<" character, we can fulfill the requirement of WebVTT format in escaping characters.


asticode:
With this change, tests fail since unescaping the `<` character in the whole file breaks the `voice` pattern (`<v Lee>What are you doing here?</v>`) 🤔

NhanNguyen700:
Yes, you are right. I correct it now.

asticode:
Thanks for the PR! ❤️

NhanNguyen700:
Thank for the merging! Could you please create a tag for it? So that I can integrate the new change with my project.

asticode:
I've created a `v0.26.0` tag 👍

PR #90, opened 2023-08-05
fix: vlc expect 10 columns

firodj:
Hi @asticode nice package. I've issue when writing to SSA/AAS file then loading it into VLC. The VLC could not render it. It seems VLC want **Text** field after 9th commas.


asticode:
How did you figure out this was the problem? it seems really weird to me 🤔  Could you send a sample file?

firodj:
Here are these files:
1. `en.ass` - original
2. `id.ass` - the `en.ass` re-saved by go-astisub. (without any modifications)
3. `fr.ass` - the `id.ass` muxed by mkvtoolnix.

[Kimetsu no Yaiba - Katanakaji no Sato-hen - 06.zip](https://github.com/asticode/go-astisub/files/12370253/Kimetsu.no.Yaiba.-.Katanakaji.no.Sato-hen.-.06.zip)

asticode:
OK thanks for sharing the test files 👍 

What you've done is not sufficient to fix this problem properly. What you need to do is add Layer as the first field but you need to make sure there's a value actually written. Here's what you need to do:

- First you need to remove your branch and create a brand new one since I had to force push on master and therefore there's a conflict with this PR
- you can remove `func (e ssaEvent) updateFormat()` 
- [here](https://github.com/asticode/go-astisub/blob/master/ssa.go#L1166) you need to make sure `i = astikit.IntPtr(0)` if `e.layer == nil` in case no layer is provided
- you can remove `formatMap` [here](https://github.com/asticode/go-astisub/blob/master/ssa.go#L1225)
- you can update `format` exactly like you did
- you can simplify this [for loop](https://github.com/asticode/go-astisub/blob/master/ssa.go#L1262) content to `events = append(events, newSSAEventFromItem(*i))`
- finally you need to fix tests

PR #92, opened 2023-08-26
fix: vlc expect 10 columns for ssa (new branch)

firodj:
continue from https://github.com/asticode/go-astisub/pull/90

### DIFF
--- a/ssa.go
+++ b/ssa.go
@@ -1150,12 +1150,10 @@ func (e *ssaEvent) string(format []string) string {
 			}
 		// Marked
 		case ssaEventFormatNameMarked:
-			if e.marked != nil {
-				if *e.marked {
-					v = "Marked=1"
-				} else {
-					v = "Marked=0"
-				}
+			if e.marked != nil && *e.marked {
+				v = "Marked=1"
+			} else {
+				v = "Marked=0"
 			}
 		// Int
 		case ssaEventFormatNameLayer, ssaEventFormatNameMarginL, ssaEventFormatNameMarginR,
@@ -1163,11 +1161,7 @@ func (e *ssaEvent) string(format []string) string {
 			var i *int
 			switch attr {
 			case ssaEventFormatNameLayer:
-				if e.layer == nil {
-					i = astikit.IntPtr(0)
-				} else {
-					i = e.layer
-				}
+				i = e.layer
 			case ssaEventFormatNameMarginL:
 				i = e.marginLeft
 			case ssaEventFormatNameMarginR:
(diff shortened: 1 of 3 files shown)


firodj:
Based on this [link](https://wiki.multimedia.cx/index.php/SubStation_Alpha), *v4.00* use `Marked` but *v4.00+* use `Layer` on the 1st column.

ssa.go
```diff
@@ -1163,7 +1163,11 @@ func (e *ssaEvent) string(format []string) string {
 			var i *int
 			switch attr {
 			case ssaEventFormatNameLayer:
-				i = e.layer
+				if e.layer == nil {
```
asticode, line 1166: Could you move this `if` line 1178 so that it "protects" all *int ?

Also, could you make that for `ssaEventFormatNameMarked` there's `Marked=0` if value is nil?

ssa.go
```diff
@@ -1253,16 +1262,23 @@ func (s Subtitles) WriteToSSA(o io.Writer) (err error) {
...
+			ssaEventFormatNameMarginV,
+			ssaEventFormatNameEffect,
+		}
+		if v4plus {
+			format[0] = ssaEventFormatNameLayer
 		}
 		var events []*ssaEvent
 		for _, i := range s.Items {
-			var e = newSSAEventFromItem(*i)
-			format = e.updateFormat(formatMap, format)
```
asticode, line 1264: Could you remove the declaration of `updateFormat()` as well?

asticode (changes requested):

ssa.go
```diff
@@ -1253,16 +1262,23 @@ func (s Subtitles) WriteToSSA(o io.Writer) (err error) {
 		var b = []byte("\n[Events]\n")
 
 		// Format
-		var formatMap = make(map[string]bool)
 		var format = []string{
```
asticode, line 1265: Also could you add the following comment above `format`?

```
We need to declare those 9 columns here otherwise VLC doesn't display subtitles properly column
```

ssa.go
firodj: sure, i've change it on the last  commit.

ssa.go
firodj: sure, i've change it on the last  commit.

ssa.go
firodj: sure, i've added it on the last  commit.

asticode: I made a typo 😱 Could you remove the `column` at the end of the comment?

ssa.go
asticode: I don't see the removal of the declaration of `updateFormat()` line 671 in your last commit 🤔

asticode (changes requested):

asticode:
Let me know whether you need a new tag 👍

PR #93, opened 2023-09-30
Support parsing srt without sequence numbers.

samuel:
There are many srt in the wild that are missing sequence numbers. It's easy to detect by looking for an empty line vs index/sequence in the line before the time. In my analysis of many srt I have not seen ones missing the empty line, but it's feasible that these also exist. This shouldn't affect parsing otherwise, however, and in my experience the missing sequence number is common enough that it's useful to handle.

### DIFF
--- a/srt_test.go
+++ b/srt_test.go
@@ -27,3 +27,22 @@ func TestSRT(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, string(c), w.String())
 }
+
+func TestSRTMissingSequence(t *testing.T) {
+	// Open
+	s, err := astisub.OpenFile("./testdata/missing-sequence-in.srt")
+	assert.NoError(t, err)
+	assertSubtitleItems(t, s)
+
+	// No subtitles to write
+	w := &bytes.Buffer{}
+	err = astisub.Subtitles{}.WriteToSRT(w)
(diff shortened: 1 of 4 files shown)


samuel:
Examples of these srt files: https://media.defcon.org/DEF%20CON%2023/DEF%20CON%2023%20video/

srt_test.go
```diff
@@ -27,3 +27,22 @@ func TestSRT(t *testing.T) {
...
+	assert.NoError(t, err)
+	assertSubtitleItems(t, s)
+
+	// No subtitles to write
+	w := &bytes.Buffer{}
+	err = astisub.Subtitles{}.WriteToSRT(w)
+	assert.EqualError(t, err, astisub.ErrNoSubtitlesToWrite.Error())
+
+	// Write
+	c, err := ioutil.ReadFile("./testdata/missing-sequence-out.srt")
```
asticode, line 43: I don't think we need to use another file here since it's the same output as `./testdata/example-out.srt` 🤔

asticode (commented):
Nice PR 👍 

I've only one comment though

srt_test.go
samuel: Great point, thanks. I'll remove it and use the existing output.

asticode:
Thanks for the PR! ❤️

asticode:

PR #95, opened 2023-10-19
Simple support for WebVTT style

NhanNguyen700:
Hi all,

Now, we will lose all of WebVTT style after reading it, process somethings and write to WebVTT format again. This PR provide a simple WebVTT style handling, since CSS support many properties, I will not parsing them all, we will just blindly store all the things in `STYLE` and rewrite them when writing WebVTT again. We will love some spaces and new lines, but it's not a big duel. Besides, when merging `subtitles2` into `subtitles1`, if there are duplicated CSS selectors, only the CSS style of `subtitles1` will be kept.

### DIFF
--- a/subtitles.go
+++ b/subtitles.go
@@ -243,10 +243,10 @@ type StyleAttributes struct {
 	WebVTTRegionAnchor   string
 	WebVTTScroll         string
 	WebVTTSize           string
+	WebVTTStyles         []string
 	WebVTTVertical       string
 	WebVTTViewportAnchor string
 	WebVTTWidth          string
-	WebVTTCSS            string
 }
 
 func (sa *StyleAttributes) propagateSSAAttributes() {}
--- a/subtitles.go
+++ b/subtitles.go
@@ -246,6 +246,7 @@ type StyleAttributes struct {
 	WebVTTVertical       string
 	WebVTTViewportAnchor string
 	WebVTTWidth          string
+	WebVTTCSS            string
 }
 
 func (sa *StyleAttributes) propagateSSAAttributes() {}
--- a/webvtt.go
+++ b/webvtt.go
@@ -3,7 +3,6 @@ package astisub
 import (
 	"bufio"
 	"bytes"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -20,10 +19,12 @@ import (
 
 // Constants
 const (
+	cssEndToken                   = "}"
 	webvttBlockNameComment        = "comment"
 	webvttBlockNameRegion         = "region"
 	webvttBlockNameStyle          = "style"
 	webvttBlockNameText           = "text"
+	webvttDefaultStyleID          = "astisub-webvtt-default-style-id"
 	webvttTimeBoundariesSeparator = " --> "
 	webvttTimestampMap            = "X-TIMESTAMP-MAP"
 )
@@ -37,6 +38,11 @@ var (
 	webVTTEscaper                      = strings.NewReplacer("&", "&amp;", "<", "&lt;")
 	webVTTUnescaper                    = strings.NewReplacer("&amp;", "&", "&lt;", "<")
 )
+
+// endOfCss returns true if the cssLines are empty or the last line is the end token
+func endOfCss(cssLines []string) bool {
+	return len(cssLines) == 0 || cssLines[len(cssLines)-1] == cssEndToken
+}
 
 // parseDurationWebVTT parses a .vtt duration
 func parseDurationWebVTT(i string) (time.Duration, error) {
@@ -110,7 +116,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	var comments []string
 	var index int
 	var timeOffset time.Duration
-	var styleLines []string
+	var webVTTStyles *StyleAttributes
 
 	for scanner.Scan() {
 		// Fetch line
@@ -125,7 +131,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 		// Empty line
 		case len(line) == 0:
 			// Reset block name
-			if blockName != webvttBlockNameStyle || len(styleLines) == 0 {
+			if blockName != webvttBlockNameStyle || webVTTStyles == nil || endOfCss(webVTTStyles.WebVTTStyles) {
 				blockName = ""
 			}
 		// Region
@@ -166,6 +172,14 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 		// Style
 		case strings.HasPrefix(line, "STYLE"):
 			blockName = webvttBlockNameStyle
+
+			if _, ok := o.Styles[webvttDefaultStyleID]; !ok {
+				webVTTStyles = &StyleAttributes{}
+				o.Styles[webvttDefaultStyleID] = &Style{
+					InlineStyle: webVTTStyles,
+					ID:          webvttDefaultStyleID,
+				}
+			}
 
 		// Time boundaries
 		case strings.Contains(line, webvttTimeBoundariesSeparator):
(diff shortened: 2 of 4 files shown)


NhanNguyen700:
Hi @asticode , I added simple support for WebVTT style, could you please take a look?

subtitles.go
```diff
@@ -246,6 +246,7 @@ type StyleAttributes struct {
 	WebVTTVertical       string
 	WebVTTViewportAnchor string
 	WebVTTWidth          string
+	WebVTTCSS            string
```
asticode, line 249: Could you rename this attribute to `WebVTTStyle` and order fields alphabetically?

webvtt.go
```diff
@@ -123,7 +125,9 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 		// Empty line
 		case len(line) == 0:
 			// Reset block name
-			blockName = ""
+			if blockName != webvttBlockNameStyle || len(styleLines) == 0 {
```
asticode, line 128: Why the ` || len(styleLines) == 0` ? 🤔

subtitles.go
asticode, line 249: I think that would be best to make this attribute a `[]string` to store each line separately

webvtt.go
```diff
@@ -109,6 +110,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	var comments []string
 	var index int
 	var timeOffset time.Duration
+	var styleLines []string
```
asticode, line 113: If `WebVTTStyle` becomes a `[]string` there's no need for this variable

webvtt.go
```diff
@@ -257,7 +262,27 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 			case webvttBlockNameComment:
 				comments = append(comments, line)
 			case webvttBlockNameStyle:
-				// TODO Do something with the style
+				styleLines = append(styleLines, line)
```
asticode, line 265: I don't really see the point of the logic below but I may be wrong.

If we make `WebVTTStyle` a `[]string`, we can `.WebVTTStyle = append(.WebVTTStyle, line)` instead.

Regarding the style id, I would create a `const webVTTDefaultStyleID = "astisub-webvtt-default-style-id"` and use it as `styleID` below.

That would do something like this:

```go
// Create style
if_, ok := o.Styles[webVTTDefaultStyleID]; !ok {
  o.Styles[webVTTDefaultStyleID] = &Style{
    InlineStyle: &StyleAttributes{},
    ID: webVTTDefaultStyleID,
  }
}

// Append line
o.Styles[webVTTDefaultStyleID].InlineStyle.WebVTTStyle = append(o.Styles[webVTTDefaultStyleID].InlineStyle.WebVTTStyle, line)
```

webvtt.go
```diff
@@ -360,11 +385,19 @@ func (s Subtitles) WriteToWebVTT(o io.Writer) (err error) {
 	var c []byte
 	c = append(c, []byte("WEBVTT\n\n")...)
 
+	for _, style := range s.Styles {
```
asticode, line 388: I don't really know whether webvtt allows several `STYLE` tokens, but with that logic, there's a risk of it. Also you need to check `InlineStyle != nil`

I would do something like this:

```go
var style []string
for _, s := range s.Styles {
  if s.InlineStyle != nil && len(s.InlineStyle.WebVTTStyle) > 0 {
    style = append(style,  s.InlineStyle.WebVTTStyle...)
  }
}
if len(style) > 0 {
  c = append(c, []byte(fmt.Sprintf("STYLE\n%s\n\n", strings.Join(style, "\n")))...)
}
```

asticode (changes requested):

subtitles.go
NhanNguyen700: Yes, I will

NhanNguyen700: :+1:

webvtt.go
NhanNguyen700: Because there could be new line in CSS, like this:
```
STYLE
::cue {
  background-image: linear-gradient(to bottom, dimgray, lightgray);

  color: papayawhip;
}
```
So if we are in the middle of a CSS, we will not reset the blockName, styleLines will be empty if we meet the end of CSS before

webvtt.go
NhanNguyen700: In fact, there can be multiple STYLE tokens in WebVTT according to the example 3 in chapter 1.3 of https://www.w3.org/TR/webvtt1/. That is why I separate the STYLE into multiple items, I don't know whether we could combine them into one without any side effect or not, so I decided to keep all STYLE tokens. Besides, it is more convenient when we merge two WebVTT files which contains different STYLE by this way, the CSS will be identified by its selector, then we can add more CSS into the merged file by comparing the selector instead of just keep the STYLE of the master file. (if your are going to merge file2 into file1, then file1 is the master file)

webvtt.go
asticode: 👍

webvtt.go
asticode: It sounds weird to me merging CSS using `Styles` id 🤔 Even if global style is split among several `STYLE` tags, there still is one and only global style global to all webvtt file, hence I'd rather store it in one pre-determined style id, as I suggested.

If we decide to tackle merging CSS then I would suggest transforming `WebVTTStyle` from `[]string` to `map[string][]string`, with the map key being the CSS selector.

NhanNguyen700: You are right, merging CSS is complicated and we should just keep the master file style, then let user know about that so they can aware instead of merging and make thing mess up.

asticode:
Let me know when I can review your last changes 👍

webvtt.go
```diff
@@ -123,7 +131,9 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 		// Empty line
 		case len(line) == 0:
 			// Reset block name
-			blockName = ""
+			if blockName != webvttBlockNameStyle || webVTTStyles == nil || endOfCss(webVTTStyles.WebVTTStyles) {
```
asticode, line 134: Does it still work if there are several CSS selectors ? 🤔 

Assume the following `Style` block:
```
STYLE
::cue(b) {
    color: peachpuff;
}

::cue(a) {
    color: white;
}
```

Does that pick up the second selector? 🤔

NhanNguyen700: Your example above violates with WebVTT format, a blank line is a sign of a completely new scope. In youe example above, we will not know when to start to parse other scopes, you can verify with this validator: https://w3c.github.io/webvtt.js/parser.html. If there is several CSS selector, the format should be
```
STYLE
::cue(b) {
    color: peachpuff;
}

STYLE
::cue(a) {
    color: white;
}
```
or 
```
STYLE
::cue(b) {
    color: peachpuff;
}
::cue(a) {
    color: white;
}
```
And my code also work with above cases

NhanNguyen700: I will add some more styles into the test input file of webvtt then.

asticode: I didn't know that 👍  Thanks!

webvtt.go
```diff
@@ -37,6 +39,11 @@ var (
 	webVTTUnescaper                    = strings.NewReplacer("&amp;", "&", "&lt;", "<")
 )
 
+// endOfCss returns true if the cssLines are empty or the last line is the end token
+func endOfCss(cssLines []string) bool {
```
asticode, line 43: It feels weird to create a function here since it's only use once below and condition is pretty simple  🤔

NhanNguyen700: The function will make the condition shorter, if we do not use this function, the condition will become:
```
if blockName != webvttBlockNameStyle || webVTTStyles == nil || len(webVTTStyles.WebVTTStyles) > 0 || webVTTStyles.WebVTTStyles[len(webVTTStyles.WebVTTStyles)-1] == cssEndToken 
```
very long and uneasy to read,

NhanNguyen700: I refer the shorten style with function since the function name state it clear what we are checking. But I am not against the long condition, so I will change the condition then.

asticode:
Tests seem to be failing, could you fix them? 🤔 

Also could you remove `cssEndToken` and use `"}"` directly in the condition? Also based on your last comment, that could be a good idea to add a comment in the condition:

```go
if blockName != webvttBlockNameStyle || webVTTStyles == nil ||
    // CSS has ended
    len(webVTTStyles.WebVTTStyles) > 0 || webVTTStyles.WebVTTStyles[len(webVTTStyles.WebVTTStyles)-1] == "}" {
```

NhanNguyen700:
Comments added and `cssEndToken` was removed

asticode:
Thanks for the PR! ❤️ 

PR #96, opened 2023-10-19
Support SRT styling and bold/underline in VTT

kloon15:
Hows this code looking like? Feedback appreciated.


srt.go
```diff
@@ -111,7 +113,88 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 			o.Items = append(o.Items, s)
 		} else {
 			// Add text
-			s.Lines = append(s.Lines, Line{Items: []LineItem{{Text: strings.TrimSpace(line)}}})
+			if l := parseTextSrt(line); len(l.Items) > 0 {
```
asticode, line 116: You can replace this with `s.Lines = append(s.Lines, parseTextSrt(strings.TrimSpace(line)).Items...)`

srt.go
```diff
@@ -111,7 +113,88 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
...
+	}
+	return
+}
+
+func parseTextSrt(i string) (o Line) {
+	// Create tokenizer
+	tr := html.NewTokenizer(strings.NewReader(i))
+
+	// Loop
+	type Styles struct {
```
asticode, line 130: No need to create this struct. You can use `bold`, `italic`, `underline` and `color` variables

srt.go
```diff
@@ -164,3 +246,67 @@ func (s Subtitles) WriteToSRT(o io.Writer) (err error) {
...
+	if b {
+		c = append(c, []byte("</b>")...)
+	}
+	if color != "" {
+		c = append(c, []byte("</font>")...)
+	}
+	return
+}
+
+func getAttribute(n *html.Token, key string) (string, bool) {
```
asticode, line 303: Could you move this function to `subtitles.go`, rename it to `htmlTokenAttribute()` and sure arguments are `(t html.Token, key string) string` ?

srt.go
```diff
@@ -111,7 +113,88 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
...
+			// Parse italic/bold/underline
+			switch token.Data {
+			case "b":
+				styles.bold = true
+			case "i":
+				styles.italic = true
+			case "u":
+				styles.underline = true
+			case "font":
+				color, _ := getAttribute(&token, "color")
```
asticode, line 172: Could you replace this with

```go
if c := htmlTokenAttribute(token, "color"); c != "" {
  color = &c
}
```

srt.go
```diff
@@ -111,7 +113,88 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
...
+				}
+			}
+		case html.TextToken:
+			if s := strings.TrimSpace(string(tr.Raw())); s != "" {
+				// Get style attribute
+				var sa *StyleAttributes
+				if styles.bold || styles.italic ||
+					styles.underline || styles.color != nil {
+					sa = &StyleAttributes{
+						TTMLColor:       styles.color,
```
asticode, line 184: Could you create `StyleAttributes`'s `SRTColor`, `SRTBold`, `SRTItalic` and `SRTUnderline`, and store the values inside them instead?

webvtt.go
```diff
@@ -290,7 +287,12 @@ func parseTextWebVTT(i string) (o Line) {
 	tr := html.NewTokenizer(strings.NewReader(i))
 
 	// Loop
-	italic := false
+	type Styles struct {
```
asticode, line 290: Same as `srt`, no need for this struct

subtitles.go
```diff
@@ -248,6 +250,8 @@ type StyleAttributes struct {
 	WebVTTWidth          string
 }
 
+func (sa *StyleAttributes) propagateSRTAttributes() {}
```
asticode, line 253: You need copy new `SRT*` attributes into proper fields such as corresponding `WebVTT*` fields. 

Same in `propagateWebVTTAttributes`, you need to do the reverse

asticode (changes requested):
Could you also fix webvtt tests and add srt/webvtt tests for bold/italic/underline/color?

You can either directly use `testdata/example-in.srt` in `TestSRT` for instance, or write a new test specifically for html tags.

srt.go
kloon15: Thats no bueno, i think u want this?
`s.Lines = append(s.Lines, parseTextSrt(line))`
I just straight up copied this part from the webvtt equivalent, should i change it there too then?

asticode: You're right, just replace `line` with `strings.TrimSpace(line)` only here, no need to update the webvtt counterpart

kloon15: Thats done inside the function like in webvtt.

asticode:
Could you remove the SSA-related commits from this PR?

I don't want to mix different things in the same PR.

Thanks

kloon15:
> Could you remove the SSA-related commits from this PR?
> 
> I don't want to mix different things in the same PR.
> 
> Thanks

I done nothing to SSA, its for supporting SRT position tags which use ASS tags so figured its easier to call em that.

asticode:
> I done nothing to SSA, its for supporting SRT position tags which use ASS tags so figured its easier to call em that. TBH im not sure what tags most players support, but position is pretty popular.

OK, then could you remove the 2 commits handling ASS tags from this PR? I'd rather have this change in a separate PR.

Thanks!

kloon15:
Sure will do that.

asticode:
It's been over a year, let me know whether this PR is still active 👍

kloon15:
No clue, I got busy with life, so idk what happened here since, whether its still relevant.

asticode:
No worries, thanks again ❤️

PR #97, opened 2023-11-13
Preserve WebVTT tags in subtitle lines

NhanNguyen700:
Hi all,

WebVTT STYLE block is CSS and it allow to write custom CSS class or tag in xml to applied to the lines of WebVTT, also, some default tag like tag i (italic), b (bold), u (underline), etc are supported too, for example:
```
WEBVTT

00:01:00.000 --> 00:02:00.000
<u><i>Italic with underline text</i></u> some extra

00:02:00.000 --> 00:03:00.000
<lang en>English here</lang> <c.yellow.bg_blue>Yellow text on blue background</c>

00:03:00.000 --> 00:04:00.000
<v Joe><c.red><i>Joe's words are red in italic</i></c>

00:04:00.000 --> 00:05:00.000
<customed_tag.class1.class2>Text here</customed_tag>
```
So, I add these change to preserve the tags in WebVTT files. This will be convenient if someday, we decide to convert styling from WebVTT to TTML or another format.

According to WebVTT format at: https://www.w3.org/TR/webvtt1/

### DIFF
--- a/subtitles.go
+++ b/subtitles.go
@@ -236,7 +236,6 @@ type StyleAttributes struct {
 	TTMLWritingMode      *string
 	TTMLZIndex           *int
 	WebVTTAlign          string
-	WebVTTItalics        bool
 	WebVTTLine           string
 	WebVTTLines          int
 	WebVTTPosition       string
@@ -244,9 +243,40 @@ type StyleAttributes struct {
 	WebVTTScroll         string
 	WebVTTSize           string
 	WebVTTStyles         []string
+	WebVTTTags           []WebVTTTag
 	WebVTTVertical       string
 	WebVTTViewportAnchor string
 	WebVTTWidth          string
+}
+
+type WebVTTTag struct {
+	Name       string
+	Annotation string
+	Classes    []string
+}
+
+func (t WebVTTTag) startTag() string {
+	if t.Name == "" {
+		return ""
+	}
+
+	s := t.Name
+	if len(t.Classes) > 0 {
+		s += "." + strings.Join(t.Classes, ".")
+	}
+
+	if t.Annotation != "" {
+		s += " " + t.Annotation
+	}
+
+	return "<" + s + ">"
+}
+
+func (t WebVTTTag) endTag() string {
+	if t.Name == "" {
+		return ""
+	}
+	return "</" + t.Name + ">"
 }
 
 func (sa *StyleAttributes) propagateSSAAttributes() {}
--- a/webvtt.go
+++ b/webvtt.go
@@ -2,10 +2,10 @@ package astisub
 
 import (
 	"bufio"
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"regexp"
 	"sort"
 	"strconv"
@@ -33,7 +33,7 @@ var (
 	bytesWebVTTItalicEndTag            = []byte("</i>")
 	bytesWebVTTItalicStartTag          = []byte("<i>")
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
-	webVTTRegexpStartTag               = regexp.MustCompile(`(<v([\.\w]*)(.+?)>)`)
+	webVTTRegexpTag                    = regexp.MustCompile(`(</*\s*([^\.\s]+)(\.[^\s/]*)*\s*([^/]*)\s*/*>)`)
 	webVTTEscaper                      = strings.NewReplacer("&", "&amp;", "<", "&lt;")
 	webVTTUnescaper                    = strings.NewReplacer("&amp;", "&", "&lt;", "<")
 )
@@ -306,12 +306,12 @@ func parseTextWebVTT(i string) (o Line) {
 	// Create tokenizer
 	tr := html.NewTokenizer(strings.NewReader(i))
 
+	webVTTTagStack := make([]WebVTTTag, 0, 16)
+
 	// Loop
-	italic := false
 	for {
 		// Get next tag
 		t := tr.Next()
-
 		// Process error
 		if err := tr.Err(); err != nil {
 			break
@@ -319,32 +319,52 @@ func parseTextWebVTT(i string) (o Line) {
 
 		switch t {
 		case html.EndTagToken:
-			// Parse italic
-			if bytes.Equal(tr.Raw(), bytesWebVTTItalicEndTag) {
-				italic = false
-				continue
+			// Pop the top of stack if we meet end tag
+			if len(webVTTTagStack) > 0 {
+				webVTTTagStack = webVTTTagStack[:len(webVTTTagStack)-1]
 			}
 		case html.StartTagToken:
-			// Parse voice name
-			if matches := webVTTRegexpStartTag.FindStringSubmatch(string(tr.Raw())); len(matches) > 3 {
-				if s := strings.TrimSpace(matches[3]); s != "" {
-					o.VoiceName = s
+			if matches := webVTTRegexpTag.FindStringSubmatch(string(tr.Raw())); len(matches) > 4 {
+				tagName := matches[2]
+
+				var classes []string
+				if matches[3] != "" {
+					classes = strings.Split(strings.Trim(matches[3], "."), ".")
 				}
-				continue
+
+				annotation := ""
+				if matches[4] != "" {
+					annotation = strings.TrimSpace(matches[4])
+				}
+
+				if tagName == "v" {
+					if o.VoiceName == "" {
+						// Only get voicename of the first <v> appears in the line
+						o.VoiceName = annotation
+					} else {
+						// TODO: do something with other <v> instead of ignoring
+						log.Printf("astisub: found another voice name %q in %q. Ignore", annotation, i)
+					}
+					continue
+				}
+
+				// Push the tag to stack
+				webVTTTagStack = append(webVTTTagStack, WebVTTTag{
+					Name:       tagName,
+					Classes:    classes,
+					Annotation: annotation,
+				})
 			}
 
-			// Parse italic
-			if bytes.Equal(tr.Raw(), bytesWebVTTItalicStartTag) {
-				italic = true
-				continue
-			}
 		case html.TextToken:
 			if s := strings.TrimSpace(string(tr.Raw())); s != "" {
 				// Get style attribute
 				var sa *StyleAttributes
-				if italic {
+				if len(webVTTTagStack) > 0 {
+					tags := make([]WebVTTTag, len(webVTTTagStack))
+					copy(tags, webVTTTagStack)
 					sa = &StyleAttributes{
-						WebVTTItalics: italic,
+						WebVTTTags: tags,
 					}
 					sa.propagateWebVTTAttributes()
 				}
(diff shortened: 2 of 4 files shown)


NhanNguyen700:
Hi @asticode , can you have a look?

webvtt.go
```diff
@@ -306,45 +305,64 @@ func parseTextWebVTT(i string) (o Line) {
 	// Create tokenizer
 	tr := html.NewTokenizer(strings.NewReader(i))
 
+	webVTTTagStack := make([]*WebVTTTag, 64)
```
asticode, line 308: Could you explain why you're using an intermediate stack with an index here? 🤔

webvtt.go
```diff
@@ -306,45 +305,64 @@ func parseTextWebVTT(i string) (o Line) {
...
-			// Parse italic
-			if bytes.Equal(tr.Raw(), bytesWebVTTItalicStartTag) {
-				italic = true
-				continue
+				annotation := ""
+				if matches[4] != "" {
+					annotation = strings.TrimSpace(matches[4])
+				}
+
+				// Only support voicename if <v> appears in the begining of the line
```
asticode, line 342: Why only support voicename if it's present at the beginning and not elsewhere? 🤔

asticode (changes requested):

webvtt.go
NhanNguyen700: As I know, go 1.13 (or even newer) do not have stack, there is proposal but it is never in a release of golang and I don't know why: https://pkg.go.dev/github.com/golang-collections/collections@v0.0.0-20130729185459-604e922904d3/stack. Thus, I choose this way to implement a simple stack. Besides, a line of WebVTT is in XML format (with the exception is `<v>` tag, the `</v>` can be deduced if the voice is applied for the whole line), so for parsing it, we need recursion because XML is a parent-children structure. To make the function simple and avoid the complexity in recursion, stack here for de-recursion

webvtt.go
NhanNguyen700: The .VoiceName was added to support converting from webvtt to ssa/ass format, and ssa/ass only support a voice name for entire line, as far as I know. However, WebVTT support multiple voice in the same line, like this:
```
Someone's saying <v Joe> So am I </v> <v John> Me too </v> <i>Joe and John talk at the same time</i>
```
And With the current behavior "John" will be the voice for ssa/ass, with this behavior no voice will be picked because both are in correct for the voice name of the line.


For preserve `<v>` tag in WebVTT format, the story is more complex, consider this example:
```
<v Ana> I think this dish is <u>fabulous</u><v> 
```
When applied the PR's new logic, the outcome is:
```
<v Ana> I think this dish is</v> <v Ana><u>fabulous</u></v>
```
I don't know if this styling is the same with the original one and looked for a way to verify, but hopeless. So I decided to postpone until I can make sure that we will now get into any trouble because of the outcome like that.

webvtt.go
asticode: I may have misunderstood the logic but that seems overcomplicated 🤔  but I may be wrong 🤔 

What about simplifying to:

```go
tags := []WebVTTTag{}
for {
  ...
  switch t {
    case html.EndTagToken:
      tags = tags[:len(tags)-1]
    case html.StartTagToken:
      // Parse tag
      tags = append(tags, WebVTTTag{
	        Name:       tagName,
	        Classes:    classes,
	        Annotation: annotation,
        })
    case html.TextToken:
      if len(tags) > 0 {
	sa = &StyleAttributes{}
        sa.tags = make([]WebVTTTag, len(tags))
	copy(sa.tags, tags)
	sa.propagateWebVTTAttributes()
      }
  }
}
```

webvtt.go
asticode: Thanks for the clarification 👍 

Instead of creating `loopCount`, could you replace the `if` condition with `if o.VoiceName == "" && tagName == "v"`?

webvtt.go
NhanNguyen700: This is also a stack, right? Just eliminate the index (the index is now actually the length of slice), I wonder if the performance is better or equal :thinking: . Let me do some benchmark then, if the performance is good enough, this should be a better approach for implementing a simple stack.

webvtt.go
NhanNguyen700: For the example that I gave above:
```
Someone's saying <v Joe> So am I </v> <v John> Me too </v> <i>Joe and John talk at the same time</i>
```
With your condition, `Joe` will be chosen, but `John` also talk something. So I think choosing `Joe` or `John` is also wrong, and eliminate the voice is a better idea? Or we need to keep one of them?  I refer choosing no one as voice name, since no matter who we choose, we will be incorrect.

webvtt.go
asticode: The behavior is similar to a stack, but I find it more readable which is better for maintainability purposes 👍

webvtt.go
asticode: Thing is, if we choose no voice, we're throwing data away and we're kind of also wrong 🤔 I'd be OK storing the first voice we find and add a TODO to process other voices.

Otherwise I guess this condition would do the trick:

```go
if  tagName == "v" {
  if o.VoiceName == "" {
    o.VoiceName = annotation
  } else {
    // Clean voice name so that we store no voice
    o.VoiceName = ""
  }
}
```

webvtt.go
NhanNguyen700: I did some benchmark and this is the result of new approach:
```
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkIngestStack-8   	   10000	    135770 ns/op	  141616 B/op	     847 allocs/op
PASS
ok  	github.com/edgeware/sbgo/cmd/ew-vodingest	1.386s
```
And the current approach in this PR:
```
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
BenchmarkIngestStack1-8   	   21097	     57454 ns/op	   33408 B/op	     696 allocs/op
PASS
ok  	github.com/edgeware/sbgo/cmd/ew-vodingest	1.806s
```
As you can see, the current approach is two time faster than the new approach, so I am going keep the current approach with some comment about push-pop from the stack.

Benchmark code:
```go
type TestStruct struct {
	field1 string
	field2 string
	field3 []string
}

func BenchmarkIngestStackNew(b *testing.B) {
	testCases := []int{63, 4, 6, 15, 3, 4, 9, 12, 8, 63, 12, 34, 55, 42, 32, 33, 45, 12, 33, 45, 63, 21, 21, 21, 23, 12, 5}
	for i := 0; i < b.N; i++ {
		for _, tc := range testCases {
			var stack []TestStruct
			for j := 0; j < tc; j++ {
				stack = append(stack, TestStruct{"a", "v", []string{"a", "b", "c"}})
			}
			for j := 0; j < tc; j++ {
				stack = stack[:len(stack)-1]
			}
		}
	}
}

func BenchmarkIngestStackOriginal(b *testing.B) {
	testCases := []int{63, 4, 6, 15, 3, 4, 9, 12, 8, 63, 12, 34, 55, 42, 32, 33, 45, 12, 33, 45, 63, 21, 21, 21, 23, 12, 5}
	for i := 0; i < b.N; i++ {
		for _, tc := range testCases {
			stack := make([]TestStruct, 64)
			index := -1
			for j := 0; j < tc; j++ {
				index++
				stack[index] = TestStruct{"a", "v", []string{"a", "b", "c"}}
			}
			for j := 0; j < tc; j++ {
				index--
			}
		}
	}
}
```

webvtt.go
NhanNguyen700: well, if we have three voice, the third voice will be chosen then :thinking: . Alright, I will choose the first voice for the voice name of the whole line, it would be good if we print some thing out to warn user that we currently ignore other voices in the the same line.

asticode: 👍

webvtt.go
asticode: Could you make a benchmark of the whole `parseTextWebVTT` function instead? I've the feeling the difference you measured is negligible compared to the whole function (e.g. regexp execution). If that's the case, I'd rather have a simpler logic to maintain than a slightly more efficient one. Besides, won't your code panic if there are more than 64 items in the stack?

NhanNguyen700: > Besides, won't your code panic if there are more than 64 items in the stack?
Yes, of course, but I think no one will make a line with 64 styling

NhanNguyen700: Actually, the new approach is still slower, but it is because of allocation of the stack whenever we push a new item into the stack. So, to mitigate the times we allocate memory, I suggest to merge two approach, like this:
```go
tags := make([]WebVTTTag, 0, 16)
for {
  ...
  switch t {
    case html.EndTagToken:
      tags = tags[:len(tags)-1]
    case html.StartTagToken:
      // Parse tag
      tags = append(tags, WebVTTTag{
	        Name:       tagName,
	        Classes:    classes,
	        Annotation: annotation,
        })
    case html.TextToken:
      if len(tags) > 0 {
	sa = &StyleAttributes{}
        sa.tags = make([]WebVTTTag, len(tags))
	copy(sa.tags, tags)
	sa.propagateWebVTTAttributes()
      }
  }
}
```
16 for initializing the length of stack is enough IMO, no one will put more than 16 styling into a line, even they do that crazy thing, this approach will not crash, just a little bit slower because allocating memory is now triggered again.

asticode: 👍 👍

subtitles.go
```diff
@@ -236,19 +236,49 @@ type StyleAttributes struct {
...
 	WebVTTWidth          string
 }
 
+type WebVTTTag struct {
+	Name       string
+	Annotation string
+	Classes    []string
+}
+
+func (t WebVTTTag) getStartTag() string {
```
asticode, line 258: I may have forgotten, but could you rename this method to `startTag`?

subtitles.go
```diff
@@ -236,19 +236,49 @@ type StyleAttributes struct {
...
+	}
+
+	if t.Annotation != "" {
+		s += " " + t.Annotation
+	}
+
+	return "<" + s + ">"
+}
+
+func (t WebVTTTag) getEndTag() string {
```
asticode, line 275: I may have forgotten, but could you rename this method to `endTag`?

asticode (changes requested):
2 final minor changes and I'll merge this PR 👍

subtitles.go
NhanNguyen700: No, I am the one who forgot about this. My apologize.

NhanNguyen700:
@asticode , thank you for your ideas! After merging the PR, would you mind creating a release tag for new changes?

asticode:
FYI I've created a `v0.26.0` tag 👍 

Thanks again for the PR ❤️

NhanNguyen700:
Hi @asticode ,
I see that you had created a tag 0.26.0, but when including it to my project, the version 0.26.0 does not include newest change. Checked with go packages page here: https://pkg.go.dev/github.com/asticode/go-astisub@v0.26.0, there is a version of 0.26.0 but it was released in late July. There is a tag conflict?

asticode:
Yeah there was a mixup, `v0.26.0` was pointing to a commit on a dev branch therefore I've overwritten hoping it would be sufficient. But it seems that it's not.

I've created a `v0.26.1` tag pointing to the last master's commit. Let me know if that fixes your issue 👍

NhanNguyen700:
Thank you! It's working now.

PR #102, opened 2024-03-20
Add support for inline timestamps

### LINKED ISSUE #99: Timestamp tag regression in 0.26.0
Inline [timestamp tags](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API#timestamp_tag) are rendered incorrectly.

The following test illustrates this issue:
```
func TestWebVTTWithTimestampTag(t *testing.T) {
	testData := `WEBVTT

	00:01:00.000 --> 00:02:00.000
	Sentence with a timestamp<00:01:02.000> in the middle`

	s, err := astisub.ReadFromWebVTT(strings.NewReader(testData))
	require.NoError(t, err)

	require.Len(t, s.Items, 1)

	b := &bytes.Buffer{}
	err = s.WriteToWebVTT(b)
	require.NoError(t, err)
	require.Equal(t, `WEBVTT

1
00:01:00.000 --> 00:02:00.000
Sentence with an timestamp<00:01:02.000> in the middle
`, b.String())
}
```

This test passes in v0.25.1 and fails in 0.26.0. 

I can't determine if this is the same bug as #94 as the original file is no longer available. I can verify that #96 does not fix this issue. 

I was able to determine that this was introduced by https://github.com/asticode/go-astisub/commit/1e3a21158104cc0c1d81dc48b8c8302e601bb466

I was able to fix this with the following diff:
```diff
diff --git a/webvtt.go b/webvtt.go
index 3b1f5e4..4f56392 100644
--- a/webvtt.go
+++ b/webvtt.go
@@ -574,7 +574,7 @@ func (li LineItem) webVTTBytes() (c []byte) {
                        c = append(c, []byte(tag.startTag())...)
                }
        }
-       c = append(c, []byte(escapeWebVTT(li.Text))...)
+       c = append(c, []byte(li.Text)...)
        if li.InlineStyle != nil {
                noTags := len(li.InlineStyle.WebVTTTags)
                for i := noTags - 1; i >= 0; i-- {
```

But of course, this breaks the `TestWebVTTEscape` test. I'm not familiar enough with the code to determine a proper fix, but happy to do a PR if you have an idea on how to address this.

WithoutPants:
Fixes #99

### DIFF
--- a/subtitles.go
+++ b/subtitles.go
@@ -426,8 +426,8 @@ func (l Line) String() string {
 // LineItem represents a formatted line item
 type LineItem struct {
 	InlineStyle *StyleAttributes
-	Style       *Style
 	StartAt     time.Duration
+	Style       *Style
 	Text        string
 }
 
--- a/subtitles.go
+++ b/subtitles.go
@@ -427,6 +427,7 @@ func (l Line) String() string {
 type LineItem struct {
 	InlineStyle *StyleAttributes
 	Style       *Style
+	StartAt     time.Duration
 	Text        string
 }
 
--- a/webvtt.go
+++ b/webvtt.go
@@ -33,8 +33,8 @@ var (
 	bytesWebVTTItalicEndTag            = []byte("</i>")
 	bytesWebVTTItalicStartTag          = []byte("<i>")
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
+	webVTTRegexpInlineTimestamp        = regexp.MustCompile(`<((?:\d{2,}:)?\d{2}:\d{2}\.\d{3})>`)
 	webVTTRegexpTag                    = regexp.MustCompile(`(</*\s*([^\.\s]+)(\.[^\s/]*)*\s*([^/]*)\s*/*>)`)
-	inlineTimestampRegexp              = regexp.MustCompile(`<((?:\d{2,}:)?\d{2}:\d{2}\.\d{3})>`)
 	webVTTEscaper                      = strings.NewReplacer("&", "&amp;", "<", "&lt;")
 	webVTTUnescaper                    = strings.NewReplacer("&amp;", "&", "&lt;", "<")
 )
@@ -358,7 +358,6 @@ func parseTextWebVTT(i string) (o Line) {
 			}
 
 		case html.TextToken:
-			s := string(tr.Raw())
 			// Get style attribute
 			var sa *StyleAttributes
 			if len(webVTTTagStack) > 0 {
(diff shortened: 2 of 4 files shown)


subtitles.go
```diff
@@ -427,6 +427,7 @@ func (l Line) String() string {
 type LineItem struct {
 	InlineStyle *StyleAttributes
 	Style       *Style
+	StartAt     time.Duration
```
asticode, line 430: Could you order fields alphabetically?

webvtt.go
```diff
@@ -34,6 +34,7 @@ var (
 	bytesWebVTTItalicStartTag          = []byte("<i>")
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
 	webVTTRegexpTag                    = regexp.MustCompile(`(</*\s*([^\.\s]+)(\.[^\s/]*)*\s*([^/]*)\s*/*>)`)
+	inlineTimestampRegexp              = regexp.MustCompile(`<((?:\d{2,}:)?\d{2}:\d{2}\.\d{3})>`)
```
asticode, line 37: Could rename `inlineTimestampRegexp` into `webVTTRegexpInlineTimestamp` and order fields alphabetically?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
-				// Get style attribute
-				var sa *StyleAttributes
-				if len(webVTTTagStack) > 0 {
-					tags := make([]WebVTTTag, len(webVTTTagStack))
-					copy(tags, webVTTTagStack)
-					sa = &StyleAttributes{
-						WebVTTTags: tags,
-					}
-					sa.propagateWebVTTAttributes()
+			s := string(tr.Raw())
```
asticode, line 361: Could avoid creating a variable here, and feed `string(tr.Raw())` in the parse function directly below?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
-				o.Items = append(o.Items, LineItem{
-					InlineStyle: sa,
-					Text:        unescapeWebVTT(s),
-				})
+				sa.propagateWebVTTAttributes()
 			}
+
+			// Append item
+			items := parseTextWebVTTText(sa, s)
+			o.Items = append(o.Items, items...)
```
asticode, line 375: Could you simplify with `o.Items = append(o.Items, parseTextWebVTTText(sa, string(tr.Raw()))...)` and update the comment to `// Append items`?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
 		}
 	}
 	return
 }
 
+func parseTextWebVTTText(sa *StyleAttributes, line string) []LineItem {
```
asticode, line 381: Could you rename to `parseTextWebVTTTextToken`?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
+	var ret []LineItem
```
asticode, line 382: Instead of creating this var here, could you make the function signature `func name(sa *StyleAttributes, line string) (ret []LineItem)`?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
+
+	// split the line by inline timestamps
+	indexes := inlineTimestampRegexp.FindAllStringSubmatchIndex(line, -1)
```
asticode, line 385: Could you first check `if len(indexes) == 0` instead and return in the if block? That will allow to remove the `else`

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
+			}
+
+			ret = append(ret, LineItem{
+				InlineStyle: sa,
+				Text:        unescapeWebVTT(s),
+				StartAt:     t,
+			})
+		}
+	} else {
+		s := strings.TrimSpace(line)
```
asticode, line 423: Since it moves to the top, could you simplify with
```go
if s := strings.TrimSpace(line); s != "" {
  return []LineItem{{
				InlineStyle: sa,
				Text:        unescapeWebVTT(s),
			}}
}
```

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
 }
 
+func parseTextWebVTTText(sa *StyleAttributes, line string) []LineItem {
+	var ret []LineItem
+
+	// split the line by inline timestamps
+	indexes := inlineTimestampRegexp.FindAllStringSubmatchIndex(line, -1)
+	if len(indexes) > 0 {
+		// get the text before the first timestamp
+		s := strings.TrimSpace(line[:indexes[0][0]])
```
asticode, line 388: Could you use this notation instead:
```go
if s := ...; s != "" {
  ...
}
```

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
+			// Parse timestamp
+			t, err := parseDurationWebVTT(ts)
+			if err != nil {
+				log.Printf("astisub: parsing webvtt duration %s failed, ignoring: %v", ts, err)
+			}
+
+			ret = append(ret, LineItem{
+				InlineStyle: sa,
+				Text:        unescapeWebVTT(s),
+				StartAt:     t,
```
asticode, line 419: Could you order fields alphabetically?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
+			if i+1 < len(indexes) {
+				endIndex = indexes[i+1][0]
+			}
+			s := strings.TrimSpace(line[match[1]:endIndex])
+			if s == "" {
+				continue
+			}
+
+			// get the timestamp
+			ts := line[match[2]:match[3]]
```
asticode, line 408: Could you avoid creating a variable here, and feed `line[match[2]:match[3]]` directly in `parseDurationWebVTT` ?

webvtt.go
```diff
@@ -357,29 +358,80 @@ func parseTextWebVTT(i string) (o Line) {
...
+			})
+		}
+
+		for i, match := range indexes {
+			// get the text between the timestamps
+			endIndex := len(line)
+			if i+1 < len(indexes) {
+				endIndex = indexes[i+1][0]
+			}
+			s := strings.TrimSpace(line[match[1]:endIndex])
```
asticode, line 402: Could you use this notation instead:
```go
if s := ...; s == "" {
  ...
}
```

asticode (commented):
Thanks for the PR ❤️ 

Only minor adjustments needed 👍

webvtt.go
WithoutPants: Ok, but `ts` is used in the error log message as well, so I'll be repeating it there.

webvtt.go
WithoutPants: `s` is used further below, so it can't use this notation.

webvtt.go
asticode: 👍

webvtt.go
asticode: 👍

asticode:
Thanks for the PR! ❤️ 

PR #103, opened 2024-04-23
added options for maintaining timestamp for vtt

Gorillarock:


Gorillarock:
@asticode may you review adding this?

PR #105, opened 2024-05-10
fix: keep sort order

nakamume:
Using stable sort to preserve original order


asticode:
Thanks for the PR ❤️ 

PR #108, opened 2024-06-18
[ttml] Rename ttml decoder to token reader and correct test data

NhanNguyen700:
Hi,

My apologize, after changing the token reader to decoder, the code does not work as expected. Or we can say that the polymorphism of Golang is not like what we have in C++, Java, etc, there is no parent and child object so I miss-understood that it will work as expected. I need to wash my brain after years of coding in traditional polymophism. And, with some coincidence, I let vscode extension to automatically edit the input ttml, that extension added some more newlines character and make the output is pretty normal to me and I am blind. My carelessness :bow:


asticode:
No worries 👍

Issue #111, opened 2024-08-29
Support configurable X-TIMESTAMP-MAP in webvtt


nakkamarra:
I'd like to be able to set an `X-TIMESTAMP-MAP` before writing to WebVTT. Maybe could store it on `Metadata` as a struct? i.e
```go
type TimestampMap struct {
    LocalTimestamp time.Duration
    PresentationTimestamp in64
}
```

nakkamarra:
Actually, now that I look through [what the webvtt reader is doing](https://github.com/asticode/go-astisub/blob/master/webvtt.go#L265) and think about it, this could just be done by directly by calculating the offset and then manually applying it i.e `subtitles.Add(calculatedOffset)` 🤔

nakkamarra:
Actually I believe this needs to be reworked either way. 

@anupcshan @asticode I don't believe we can always just apply the timeOffset from the `X-TIMESTAMP-MAP` like we are doing [here](https://github.com/asticode/go-astisub/pull/63/files#diff-61991c3e38e65a683d9553a855b3ff61f8bb29ed780cd61d2e8f19622b899b24R270-R272), because if we want to be able to support writing the `X-TIMESTAMP-MAP` or manually applying the offset defined by it like my comment above suggests, we would have to consider the case that the current `*Subtitles` have been parsed from a WebVTT and the correction has already been applied. In this scenario we will have cue times that have been shifted by the offset, plus a header or `Add(...)` call applying the same offset again resulting in 2x the offset.

nakkamarra:
[Here](https://github.com/asticode/go-astisub/pull/112/files#diff-7b52b6599ec9e123da9db19f3cec56292f1c60ae28d75d5ee59613d60c17127dR135-R137)'s an example of the usage in a PR I've created, the usage still supports both cases I think, but in @anupcshan 's case where you want to write a WebVTT without the header and the time offset applied, you can do so with a call to `Add()`

nakkamarra:
Sorry for the annoying bump @asticode @anupcshan but any thoughts on this issue / PR?

asticode:
No worries, I have seen your PR but didn't have much time to review it. I'll try to do it next week 👍

PR #112, opened 2024-08-30
fix: Add support for X-TIMESTAMP-MAP r/w persistence

nakkamarra:
As I mentioned in [this issue here](https://github.com/asticode/go-astisub/issues/111), I'd like to be able to specify an `X-TIMESTAMP-MAP` based on some external factors in my application which uses astisub. However, as part of this, I think that the WebVTT reader automatically applying the time offset defined therein will cause duplicate applications of the time offset in the case where a WebVTT with an `X-TIMESTAMP-MAP` is read, manipulated, then also written back out. 

So this PR adds a struct for storing the parsed `X-TIMESTAMP-MAP` data on the `Subtitles`' Metadata field, that way it can be manually applied to the subtitles if necessary, or left as is and written back to WebVTT files so that some upstream client can apply the time correction. Likewise, this also allows manually configuring a `X-TIMESTAMP-MAP` to be written to a WebVTT file when it previously did not exist there, the use-case I originally needed to solve for.

### DIFF
--- a/subtitles.go
+++ b/subtitles.go
@@ -392,6 +392,24 @@ type Metadata struct {
 	STLTranslatorName                                   string
 	Title                                               string
 	TTMLCopyright                                       string
+	WebVTTXTimestampMap                                 *TimestampMap
+}
+
+// TimestampMap is a structure for storing timestamps for WEBVTT's
+// X-TIMESTAMP-MAP feature commonly used for syncing cue times with
+// MPEG-TS streams.
+type TimestampMap struct {
+	LocalTimestamp        time.Duration
+	PresentationTimestamp int64
+}
+
+// Offset calculates and returns the time offset described by the
+// timestamp map.
+func (t *TimestampMap) Offset() time.Duration {
+	if t == nil {
+		return 0
+	}
+	return time.Duration(t.PresentationTimestamp)*time.Second/90000 - t.LocalTimestamp
 }
 
 // Region represents a subtitle's region
--- a/subtitles.go
+++ b/subtitles.go
@@ -392,24 +392,7 @@ type Metadata struct {
 	STLTranslatorName                                   string
 	Title                                               string
 	TTMLCopyright                                       string
-	WebVTTXTimestampMap                                 *TimestampMap
-}
-
-// TimestampMap is a structure for storing timestamps for WEBVTT's
-// X-TIMESTAMP-MAP feature commonly used for syncing cue times with
-// MPEG-TS streams.
-type TimestampMap struct {
-	LocalTimestamp        time.Duration
-	PresentationTimestamp int64
-}
-
-// Offset calculates and returns the time offset described by the
-// timestamp map.
-func (t *TimestampMap) Offset() time.Duration {
-	if t == nil {
-		return 0
-	}
-	return time.Duration(t.PresentationTimestamp)*time.Second/90000 - t.LocalTimestamp
+	WebVTTTimestampMap                                  *WebVTTTimestampMap
 }
 
 // Region represents a subtitle's region
--- a/webvtt.go
+++ b/webvtt.go
@@ -66,7 +66,7 @@ func (t *WebVTTTimestampMap) Offset() time.Duration {
 func (t *WebVTTTimestampMap) String() string {
 	mpegts := fmt.Sprintf("MPEGTS:%d", t.MpegTS)
 	local := fmt.Sprintf("LOCAL:%s", formatDurationWebVTT(t.Local))
-	return fmt.Sprintf("%s=%s,%s", webvttTimestampMapHeader, mpegts, local)
+	return fmt.Sprintf("%s=%s,%s", webvttTimestampMapHeader, local, mpegts)
 }
 
 // https://tools.ietf.org/html/rfc8216#section-3.5
--- a/webvtt.go
+++ b/webvtt.go
@@ -25,7 +25,7 @@ const (
 	webvttBlockNameText           = "text"
 	webvttDefaultStyleID          = "astisub-webvtt-default-style-id"
 	webvttTimeBoundariesSeparator = " --> "
-	webvttTimestampMap            = "X-TIMESTAMP-MAP"
+	webvttTimestampMapHeader      = "X-TIMESTAMP-MAP"
 )
 
 // Vars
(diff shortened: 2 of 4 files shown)


subtitles.go
```diff
@@ -392,6 +392,24 @@ type Metadata struct {
 	STLTranslatorName                                   string
 	Title                                               string
 	TTMLCopyright                                       string
+	WebVTTXTimestampMap                                 *TimestampMap
+}
+
+// TimestampMap is a structure for storing timestamps for WEBVTT's
+// X-TIMESTAMP-MAP feature commonly used for syncing cue times with
+// MPEG-TS streams.
+type TimestampMap struct {
```
asticode, line 401: Could you rename this to `WebVTTTimestampMap` and move it to `webvtt.go` above `func parseWebVTTTimestampMap()`?

webvtt.go
```diff
@@ -48,7 +48,7 @@ func parseDurationWebVTT(i string) (time.Duration, error) {
 // Eg., `X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:900000` => 10s
 //
 //	`X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:180000` => 2s
-func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error) {
+func parseTimestampMapWebVTT(line string) (timestampMap *TimestampMap, err error) {
```
asticode, line 51: Could you rename the method to `parseWebVTTTimestampMap`?

webvtt.go
```diff
@@ -58,10 +58,12 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
 
 	var local time.Duration
 	var mpegts int64
+	timestampMap = new(TimestampMap)
```
asticode, line 61: Could you remove this?

webvtt.go
```diff
@@ -58,10 +58,12 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
...
 	for _, split := range strings.Split(right, ",") {
 		splits := strings.SplitN(split, ":", 2)
 		if len(splits) <= 1 {
 			err = fmt.Errorf("astisub: invalid X-TIMESTAMP-MAP, part %q didn't contain ':'", right)
+			timestampMap = nil
```
asticode, line 66: Could you remove this?

webvtt.go
```diff
@@ -70,18 +72,21 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
 			local, err = parseDurationWebVTT(splits[1])
 			if err != nil {
 				err = fmt.Errorf("astisub: parsing webvtt duration failed: %w", err)
+				timestampMap = nil
```
asticode, line 75: Could you remove this?

webvtt.go
```diff
@@ -70,18 +72,21 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
...
 				return
 			}
+			timestampMap.LocalTimestamp = local
```
asticode, line 78: Could you remove this?

webvtt.go
```diff
@@ -70,18 +72,21 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
...
 		case "mpegts":
 			mpegts, err = strconv.ParseInt(splits[1], 10, 0)
 			if err != nil {
 				err = fmt.Errorf("astisub: parsing int %s failed: %w", splits[1], err)
+				timestampMap = nil
```
asticode, line 83: Could you remove this?

webvtt.go
```diff
@@ -70,18 +72,21 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
...
 				return
 			}
+			timestampMap.PresentationTimestamp = mpegts
```
asticode, line 86: Could you remove this?

subtitles.go
```diff
@@ -392,6 +392,24 @@ type Metadata struct {
...
+	LocalTimestamp        time.Duration
```
asticode, line 402: Could you rename this field to `Local`

subtitles.go
```diff
@@ -392,6 +392,24 @@ type Metadata struct {
...
+	PresentationTimestamp int64
```
asticode, line 403: Coud you rename this field ot `MpegTS`?

webvtt.go
```diff
@@ -70,18 +72,21 @@ func parseTimestampMapWebVTT(line string) (timeOffset time.Duration, err error)
...
 		}
 	}
 
-	timeOffset = time.Duration(mpegts)*time.Second/90000 - local
```
asticode, line 84: Could you add 
```go
timestampMap = &WebVTTTimestampMap{
  Local: local,
  MpegTS: mpegts,
}
```

subtitles.go
```diff
@@ -392,6 +392,24 @@ type Metadata struct {
 	STLTranslatorName                                   string
 	Title                                               string
 	TTMLCopyright                                       string
+	WebVTTXTimestampMap                                 *TimestampMap
```
asticode, line 395: Could you rename this field to `WebVTTTimestampMap`?

webvtt.go
```diff
@@ -440,7 +445,19 @@ func (s Subtitles) WriteToWebVTT(o io.Writer) (err error) {
...
 	// Add header
 	var c []byte
-	c = append(c, []byte("WEBVTT\n\n")...)
+	c = append(c, []byte("WEBVTT")...)
+
+	// Write X-TIMESTAMP-MAP if set
+	if s.Metadata != nil {
+		if s.Metadata.WebVTTXTimestampMap != nil {
+			c = append(c, []byte("\n")...)
+			mpegts := fmt.Sprintf("MPEGTS:%d", s.Metadata.WebVTTXTimestampMap.PresentationTimestamp)
```
asticode, line 454: Could you create a `func (m *WebVTTTimestampMap) String() string` that would return the header string instead, and use this function here?

That would only leave `c = append(c, append([]byte("\n"), []byte(s.Metadata.WebVTTXTimestampMap.String())...)...)` here

webvtt_internal_test.go
```diff
@@ -107,7 +107,8 @@ func TestTimestampMap(t *testing.T) {
 		},
 	} {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
-			offset, err := parseTimestampMapWebVTT(c.line)
+			timestampMap, err := parseTimestampMapWebVTT(c.line)
+			offset := timestampMap.Offset()
```
asticode, line 111: Can you merge the 2 lines into `assert.Equal(t, c.expectedOffset, timestampMap.Offset())`?

webvtt_internal_test.go
```diff
@@ -107,7 +107,8 @@ func TestTimestampMap(t *testing.T) {
...
 			assert.Equal(t, c.expectedOffset, offset)
```
asticode, line 112: Also can you add `assert.Equal(t, c.line, timestampMap.String())`?

webvtt_test.go
```diff
@@ -121,6 +121,25 @@ func TestWebVTTWithTimestampMap(t *testing.T) {
 	err = s.WriteToWebVTT(b)
 	assert.NoError(t, err)
 	assert.Equal(t, `WEBVTT
+X-TIMESTAMP-MAP=MPEGTS:180000,LOCAL:00:00:00.000
```
asticode, line 124: I don't think we need to test the `.Add()` method here 🤔 

I would only assert the first item's start at on the first webvtt read body, and I would also assert the timestamp map header is written properly.

asticode (changes requested):
Sorry for the delay !

webvtt.go
nakkamarra: Can you explain the reasoning behind that? I'd rather return a nil + error at this point of the function since parsing failed and a non nil `&WebVTTTimestampMap` will have default values for int64 and time.Duration for both fields implying it was in some way valid.

webvtt_test.go
nakkamarra: This was a pre-existing test that originally validated that the timestamp map's offset was properly applied if the parsed WebVTT contained a map. When my changes broke the test I figured I'd just replicate the behavior that was already there with the new code. I can remove this test if you'd like.

webvtt.go
nakkamarra: Oh I see. You're leaning on the naked returns. They can be a bit confusing for me. It's a feature of golang I'm not the biggest fan of because I feel it tends to make the code a bit harder for me to follow

webvtt_test.go
nakkamarra: Adjusted, thanks 👍

asticode:
Thanks for the PR ! ❤️ 

nakkamarra:
> Thanks for the PR ! ❤️
> 
> Let me know whether you need a tag 👍

Hey yes, a new tag would be great thank you!

asticode:
FYI I've created a `v0.27.0` 👍

Issue #114, opened 2024-10-02
Unclear opinion about encodings


nakkamarra:
Currently, there are places where astisub behaves in bizarres ways when passed a file that is _not_ UTF-8 encoded. 

For example, of the SRT and WebVTT readers, neither seem to explicitly error in this case, but all of the parsing logic doesn't work correctly since the bytes are not represented the same in the source code and the file that is being parsed (i.e UTF16 has the extra `\x00` bits per bytes so we can't just check stringA == stringB since there's padding bytes). 

So I guess my question is then does this library aim to support these other encodings, or only aims to support the by far more popular UTF-8? There is a notion in the codebase of the bytes order marks that are used by non UTF-8 encodings, but they seem to just be always stripped out and not actually used to check payload encoding. Personally,I think that is perfectly fine thing to _not support_ non-UTF-8 formats, but in that case I do think this codebase might want to make a few changes to check the encoding of what it is being passed i.e a `utf.Valid(...)` call on the bytes passed in or something like that to distinctly clarify what is and what isn't supported.

asticode:
To be honest I don't really grasp what would need to be added to handle UTF-16 encoding for instance 🤔 

But anyhow, here's what I think:

- since this lib can only handle UTF-8, I do agree that it should return an error on reading if the content is not UTF-8
- I'm not against adding compatibility with non UTF-8, it will just have to go through PRs

nakkamarra:
@asticode So UTF-16 encoding encodes code points as 16 bit vs UTF-8's code points being 8-bit, so one character will be two bytes vs one byte, so you can imagine a character represented in UTF-8 by the hex value `0x01`, it's value in UTF-16 would be `0x01 0x00` big endian or `0x00 0x01` little endian, the BOM in this case dictates the endianness of the encoding (whereas in UTF-8 it's kind of just a placeholder, it is sometimes considered optional). 

I think that UTF-8 being so predominant on the web and since UTF-16 makes up such a small subset of cases, that it might not be worth supporting other encodings but yeah, validating that something is UTF-8 before trying to read the bytes using parser x would probably be good. I can make a PR

asticode:
A PR would be great ! 👍

nakkamarra:
OK, I've got a working idea of how I want to do in [this draft](https://github.com/asticode/go-astisub/pull/115), let me know what you think. I am unsure about the other file formats supported here and if they also expect input to be UTF-8, but probably safe to do to those as well if you agree.

[...]

PR #115, opened 2024-10-03
Enforce UTF-8 encoding for input bytes

nakkamarra:
As discussed in #114, it would be better if we tried not to read bytes in other encoding if possible, as the parsers will become confused since they are generally using strings from a go source file (UTF-8) to compare against the input bytes which may or may not be UTF-8.

### DIFF
--- a/srt.go
+++ b/srt.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 )
 
 // Constants
@@ -42,6 +43,10 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 	for scanner.Scan() {
 		// Fetch line
 		line = strings.TrimSpace(scanner.Text())
+		if !utf8.ValidString(line) {
+			err = fmt.Errorf("astisub: bytes are not valid utf-8")
+			return
+		}
 		lineNum++
 
 		// Remove BOM header
--- a/webvtt.go
+++ b/webvtt.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"golang.org/x/net/html"
 )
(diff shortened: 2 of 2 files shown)


srt_test.go
```diff
@@ -46,3 +46,9 @@ func TestSRTMissingSequence(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, string(c), w.String())
 }
+
+func BenchmarkOpenSRT(b *testing.B) {
```
asticode, line 50: Could you remove this benchmark and instead add a test to validate reading non-utf8 `srt` returns an error?

webvtt_test.go
```diff
@@ -227,3 +227,9 @@ func TestWebVTTTags(t *testing.T) {
 Text with a <00:06:30.000>timestamp in the middle
 `, b.String())
 }
+
+func BenchmarkOpenWebVTT(b *testing.B) {
```
asticode, line 231: Could you remove this benchmark and instead add a test to validate reading non-utf8 `webvtt` returns an error (in both header and afterwards)?

asticode (changes requested):

srt_test.go
```diff
@@ -46,3 +47,16 @@ func TestSRTMissingSequence(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, string(c), w.String())
 }
+
+func TestNonUTF8SRT(t *testing.T) {
+	_, err := astisub.OpenFile("./testdata/example-in-non-utf8.srt")
+	assert.Error(t, err)
+}
+
+func BenchmarkOpenSRT(b *testing.B) {
```
asticode, line 56: Could you remove this benchmark?

webvtt_test.go
```diff
@@ -227,3 +233,11 @@ func TestWebVTTTags(t *testing.T) {
...
```
asticode, line 237: Could you remove this benchmark?

asticode (changes requested):
Once the benchmarks will be removed, I'll merge the PR 👍

nakkamarra:
> Once the benchmarks will be removed, I'll merge the PR 👍

Cool, removed. Thanks @asticode!

nakkamarra:
@asticode A tag would be great thanks!

asticode:
FYI I've created a `v0.29.0` tag 👍

PR #122, opened 2024-12-13
SRT HTML Entities

justin-taylor:
This PR solves an issue where srt files output to webvtt are double encoded. This also updates the current webvtt escaping to be generalized with all HTML entities. For example, currently an srt file containing an `&nbsp;` will output to `&amp;nbsp;` when being written to vtt.

### DIFF
--- a/srt.go
+++ b/srt.go
@@ -202,10 +202,9 @@ func parseTextSrt(i string) (o Line) {
 				}
 
 				// Append item
-				s = unescapeWebVTT(s)
 				o.Items = append(o.Items, LineItem{
 					InlineStyle: sa,
-					Text:        s,
+					Text:        unescapeHTML(s),
 				})
 			}
 		}
@@ -306,7 +305,7 @@ func (li LineItem) srtBytes() (c []byte) {
 	if pos != 0 {
 		c = append(c, []byte(fmt.Sprintf(`{\an%d}`, pos))...)
 	}
-	c = append(c, []byte(escapeWebVTT(li.Text))...)
+	c = append(c, []byte(escapeHTML(li.Text))...)
 	if u {
 		c = append(c, []byte("</u>")...)
 	}
--- a/srt.go
+++ b/srt.go
@@ -202,6 +202,7 @@ func parseTextSrt(i string) (o Line) {
 				}
 
 				// Append item
+				s = unescapeWebVTT(s)
 				o.Items = append(o.Items, LineItem{
 					InlineStyle: sa,
 					Text:        s,
@@ -305,7 +306,7 @@ func (li LineItem) srtBytes() (c []byte) {
 	if pos != 0 {
 		c = append(c, []byte(fmt.Sprintf(`{\an%d}`, pos))...)
 	}
-	c = append(c, []byte(li.Text)...)
+	c = append(c, []byte(escapeWebVTT(li.Text))...)
 	if u {
 		c = append(c, []byte("</u>")...)
 	}
--- a/srt_test.go
+++ b/srt_test.go
@@ -127,24 +127,3 @@ func TestSRTStyled(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, string(c), w.String())
 }
-
-func TestSRTHTMLEntity(t *testing.T) {
-	s, err := astisub.OpenFile("./testdata/example-in-html-entity.srt")
-	assert.NoError(t, err)
-
-	//Write to srt
-	w := &bytes.Buffer{}
-	c, err := os.ReadFile("./testdata/example-in-html-entity.srt")
-	assert.NoError(t, err)
-	err = s.WriteToSRT(w)
-	assert.NoError(t, err)
-	assert.Equal(t, string(c), w.String())
-
-	//Write WebVTT
-	w = &bytes.Buffer{}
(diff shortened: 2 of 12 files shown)


srt.go
```diff
@@ -202,6 +202,7 @@ func parseTextSrt(i string) (o Line) {
 				}
 
 				// Append item
+				s = unescapeWebVTT(s)
```
asticode, line 205: Could you:
- move `webVTTEscaper` and `webVTTUnescaper` to `subtitles.go`
- rename them to `htmlEscaper` and `htmlUnescaper`
- move `escapeWebVTT` and `unescapeWebVTT` to `subtitles.go`
- rename them `escapeHTML` and `unescapeHTML`

asticode, line 205: Instead of adding this new line, could you unescape directly below like `Text: unescapeHTML(s),`?

webvtt.go
```diff
@@ -36,8 +36,8 @@ var (
 	bytesWebVTTTimeBoundariesSeparator = []byte(webvttTimeBoundariesSeparator)
 	webVTTRegexpInlineTimestamp        = regexp.MustCompile(`<((?:\d{2,}:)?\d{2}:\d{2}\.\d{3})>`)
 	webVTTRegexpTag                    = regexp.MustCompile(`(</*\s*([^\.\s]+)(\.[^\s/]*)*\s*([^/]*)\s*/*>)`)
-	webVTTEscaper                      = strings.NewReplacer("&", "&amp;", "<", "&lt;")
-	webVTTUnescaper                    = strings.NewReplacer("&amp;", "&", "&lt;", "<")
+	webVTTEscaper                      = strings.NewReplacer("astisub.amp", "&")
```
asticode, line 39: I don't really like the idea of using this `astisub.amp` string to mess with html escape/unescape 🤔 

Wouldn't adding `"\u00A0", "&nbsp;"` to `htmlEscaper` and `"&nbsp;", "\u00A0""` to `htmlUnescaper` do the trick? 🤔 ([\u00A0](https://www.compart.com/en/unicode/U+00A0))

webvtt_test.go
```diff
@@ -158,8 +158,8 @@ func TestWebVTTEscape(t *testing.T) {
 	require.NoError(t, err)
```
asticode, line 158: Could you add a `&nbsp;` above to test this case as well?

srt_test.go
```diff
@@ -127,3 +127,24 @@ func TestSRTStyled(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, string(c), w.String())
 }
+
+func TestSRTHTMLEntity(t *testing.T) {
```
asticode, line 131: Could you :
- move this test to `subtitles_test.go`
- rename it to `TestHTMLEntities`
- rename `example-in-html-entity.srt` to `example-in-html-entities.srt` and duplicate it to `example-out-html-entities.srt`
- rename `example-out-html-entity.vtt` to `example-out-html-entities.vtt` and duplicate it to `example-in-html-entities.vtt`
- make sure all those files have `&amp;`, `&lt;` and `&nbsp;` in them
- in the test do the following
  - open `in.srt`
  - assert lines like [here](https://github.com/asticode/go-astisub/blob/master/webvtt_test.go#L161)
  - assert write to srt with `out.srt` file and write to webvtt with `out.vtt`
  - open `in.vtt`
  - assert lines like [here](https://github.com/asticode/go-astisub/blob/master/webvtt_test.go#L161)
  - assert write to srt with `out.srt` file and write to webvtt with `out.vtt`
- remove `TestWebVTTEscape` in `webvtt_test.go`

asticode (changes requested):

webvtt.go
justin-taylor: This solves for all html entities (e.g. `&euro;` `&quot;` ...) rather than keeping a short list of acceptable entities. I tried just using go's `html.EscapeString`, however that leads to converting everything in the output. So, things like single quotes are forced to `&#39;` double quotes go to `&#34;`. Not sure if that is acceptable for this library.

srt.go
justin-taylor: Done

webvtt_test.go
justin-taylor: Test case removed in favor of test in subtitles.go

srt_test.go
justin-taylor: Done. Test reflects the current `astistub.amp` implementation, but can be updated once that discussion is resolved.

webvtt.go
asticode: The problem is that when converting to `ssa`, `stl` or `teletext` there will be `astisub.amp` as well in the output. What is parsed from any subtitle format and written in the `Subtitles` struct should not be format-specific and should be usable by all formats. Therefore we need to transform the html entities into "neutral" characters that express themselves properly and are usable by all formats: `&amp;` can become `&`, `&lt;` `<`, `&euro;` `€`, `&quot;` `"`, etc. Unfortunately, I don't see another way than having a short list of acceptable entities 🤔

justin-taylor: Understood. I updated my changes back to using a short list of acceptable entities with nbsp added. I might look into a larger solution for this later.

asticode:
Thanks for the PR! ❤️

PR #123, opened 2024-12-15
Allow for styles to persist over lines

justin-taylor:
This PR solves for an issue where styles are not maintained across lines when unterminated. For example:
```
<i>The man in black fled across the desert
and the gunslinger followed</i>
```

Currently, this results in:
```
<i>The man in black fled across the desert</i>
and the gunslinger followed
```

This PR updates the results to:
```
<i>The man in black fled across the desert</i>
<i>and the gunslinger followed</i>
```


srt.go
```diff
@@ -127,24 +133,18 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 }
 
 // parseTextSrt parses the input line to fill the Line
-func parseTextSrt(i string) (o Line) {
+func parseTextSrt(i string, styles StyleAttributes) (Line, StyleAttributes) {
```
asticode, line 136: Could you change the function signature to `func parseTextSrt(i string, sa *StyleAttributes) (o Line)` instead? That way you can update the pointer directly 👍

srt.go
```diff
@@ -127,24 +133,18 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
...
 	// special handling needed for empty line
-	if strings.TrimSpace(i) == "" {
+	o := Line{}
```
asticode, line 138: Why remove the `TrimSpace()`? 🤔

srt.go
```diff
@@ -164,39 +164,38 @@ func parseTextSrt(i string) (o Line) {
...
-					color = c
+					styles.SRTColor = c
 				}
 			}
 		case html.TextToken:
 			if s := strings.TrimSpace(raw); s != "" {
 				// Get style attribute
 				var sa *StyleAttributes
-				if bold || italic || underline || color != nil || pos != 0 {
+				if styles.SRTBold || styles.SRTColor != nil || styles.SRTItalics || styles.SRTUnderline {
```
asticode, line 193: Why did you remove the handling of `position` here? 🤔

srt.go
```diff
@@ -164,39 +164,38 @@ func parseTextSrt(i string) (o Line) {
...
 					sa = &StyleAttributes{
```
asticode, line 194: You should be able to use `sa` instead of creating this here

srt_test.go
```diff
@@ -80,36 +80,47 @@ func TestSRTStyled(t *testing.T) {
...
 	assert.Equal(t, "'The time is 7:35.'", s.Items[5].Lines[1].String())
+	assert.Equal(t, "Test with multi line italics", s.Items[6].Lines[0].String())
+	assert.Equal(t, "Terminated on the next line", s.Items[6].Lines[1].String())
+	assert.Equal(t, "Unterminated styles", s.Items[7].Lines[0].String())
+	assert.Equal(t, "Do no fall to the next item", s.Items[8].Lines[0].String())
 
 	// assert the styles of the items
-	assert.Len(t, s.Items, 6)
 	assert.Equal(t, "#00ff00", *s.Items[0].Lines[0].Items[0].InlineStyle.SRTColor)
-	assert.Zero(t, s.Items[0].Lines[0].Items[0].InlineStyle.SRTPosition)
```
asticode, line 87: You shouldn't have to remove all asserts on `position` 🤔

webvtt.go
```diff
@@ -340,11 +345,10 @@ func unescapeWebVTT(i string) string {
 }
 
 // parseTextWebVTT parses the input line to fill the Line
-func parseTextWebVTT(i string) (o Line) {
+func parseTextWebVTT(i string, webVTTTagStack []WebVTTTag) (Line, []WebVTTTag) {
```
asticode, line 348: Like for `srt`, could you replace with ` func parseTextWebVTT(i string, tags *[]WebVTTTag) (o Line)` instead?

webvtt.go
```diff
@@ -340,11 +345,10 @@ func unescapeWebVTT(i string) string {
...
 	// Create tokenizer
 	tr := html.NewTokenizer(strings.NewReader(i))
-
-	webVTTTagStack := make([]WebVTTTag, 0, 16)
+	o := Line{}
```
asticode, line 351: No need for `o := Line{}`

webvtt.go
```diff
@@ -169,6 +170,9 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 				strings.HasSuffix(webVTTStyles.WebVTTStyles[len(webVTTStyles.WebVTTStyles)-1], "}") {
 				blockName = ""
 			}
+			// Reset tag stack
```
asticode, line 173: Can you add a new line before the comment and change the comment to `// Reset tags`?

webvtt.go
```diff
@@ -123,6 +123,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	var scanner = bufio.NewScanner(i)
 	var line string
 	var lineNum int
+	webVTTTagStack := make([]WebVTTTag, 0, 16)
```
asticode, line 126: Can you rename this variable to `tags`?

webvtt.go
```diff
@@ -340,11 +345,10 @@ func unescapeWebVTT(i string) string {
 }
 
 // parseTextWebVTT parses the input line to fill the Line
-func parseTextWebVTT(i string) (o Line) {
+func parseTextWebVTT(i string, webVTTTagStack []WebVTTTag) (Line, []WebVTTTag) {
 	// Create tokenizer
```
asticode, line 349: You'll need to replace `webVTTTagStack` with `*tags` below

webvtt_internal_test.go
```diff
@@ -13,7 +13,7 @@ func TestParseTextWebVTT(t *testing.T) {
 	t.Run("When both voice tags are available", func(t *testing.T) {
 		testData := `<v Bob>Correct tag</v>`
 
-		s := parseTextWebVTT(testData)
+		s, _ := parseTextWebVTT(testData, make([]WebVTTTag, 0, 16))
```
asticode, line 16: The best would be to handle `nil` tags in `parseTextWebVTT` so that you can put `nil` instead of `make([]WebVTTTag, 0, 16)` here

asticode (changes requested):

srt.go
justin-taylor: The input has already been trimmed on line 48, so this is unnecessary, similar to `parseTextWebVTT`.

justin-taylor:
@asticode I believe all comments have been addressed. For webvtt I converted to passing a `*StyleAttribute` instead of the tag stack as you can't modify the length of a slice ptr. This allows makes it match up closer to the srt parsing. Let me know if anything else needs updating.

srt.go
asticode: I do agree but we may add some more processing on the `line` variable in the future and I want to make sure that `parseTextWebVTT` does not treat spaces only as anything but an empty line 👍

webvtt.go
```diff
@@ -143,7 +143,7 @@ func ReadFromWebVTT(i io.Reader) (o *Subtitles, err error) {
 	var blockName string
 	var comments []string
 	var index int
-	var webVTTStyles *StyleAttributes
+	var webVTTStyles = &StyleAttributes{}
```
asticode, line 146: Could you rename `webVTTStyles` to `sa`?

asticode:
> For webvtt I converted to passing a *StyleAttribute instead of the tag stack as you can't modify the length of a slice ptr.

Good idea 👍 

Additionally to the 2 last comments needed to be addressed, could you rebase this PR with `master` now your other PR has been merged?

justin-taylor:
Rebase, var rename, and re-add trimspace done. Let me know if there is anything else, otherwise can a new tag be made once this pr is merged 🙏?

asticode:
Thanks again for both PRs ❤️ 

I've created the `v0.31.0` tag 👍

PR #124, opened 2024-12-17
Allow for more lenient duration parsing

justin-taylor:
This PR allows for somewhat more lenient parsing of the timestamps for srt and vtt. For both srt and vtt this allows for timestamps to use the separator without the surrounding spaces. For example:

```
00:00:01.876-->00:0:03.390
```

For vtt, this PR also allows for parsing timestamps where the inline styles are separated by tabs `\t`.

### DIFF
--- a/srt.go
+++ b/srt.go
@@ -14,12 +14,12 @@ import (
 
 // Constants
 const (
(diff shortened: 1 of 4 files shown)


asticode:
Thanks for the PR! ❤️

Issue #118, opened 2024-10-09
Implement custom `bufio.SplitFunc` for WebVTT


nakkamarra:
From an [excerpt](https://www.w3.org/TR/webvtt1/#webvtt-parser-algorithm) of the WebVTT spec regarding the WebVTT parsing algorithm:

> Let input be the string being parsed, after conversion to Unicode, and with the following transformations applied:
>    Replace all U+0000 NULL characters by U+FFFD REPLACEMENT CHARACTERs.
>
>    Replace each U+000D CARRIAGE RETURN U+000A LINE FEED (CRLF) character pair by a single U+000A LINE FEED (LF) character.
>
>   Replace all remaining U+000D CARRIAGE RETURN characters by U+000A LINE FEED (LF) characters.

I don't believe the current astisub WebVTT parsing logic is covering this part of the WebVTT parsing algorithm correctly. The `bufio.Scanner` is using the default which is [ScanLines](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/bufio/scan.go;l=355) and I'm not sure this is handling the WebVTT parsing correctly (i.e a file with _exclusively_ `\r` terminations will not be replaced and thus not parsed properly). So I'm thinking maybe implementing a new `bufio.SplitFunc` and feeding it to the `scanner.Split(...)`

This somewhat ties into another issue I am currently trying to make a PR to solve #110, where I'd like to:

1. read the first 6 bytes and make sure they are `WEBVTT`
2. read the remainder of the content up until the first occurrence of two successive line terminations
3. write that content into a new field of the Subtitles struct that has each line

Thoughts? Maybe there's a better way for me to achieve what I'm trying to do.

asticode:
From what I understand, there are 2 problems you're trying to fix:

1) handle `\r`
2) handle optional data after `WEBVTT` keyword

Regarding 1), implementing a custom `bufio.SplitFunc` capable of parsing `\n\r` seems like a good idea
Regarding 2), I would update [this logic](https://github.com/asticode/go-astisub/blob/master/webvtt.go#L127-L139), trimming BOMBytes prefix, checking the header is there, scanning until 2 empty lines have been found and storing its value in a `Subtitles.Comments` attribute

PR #125, opened 2024-12-17
Add custom split logic for scanner

justin-taylor:
This PR provides a custom `bufio.SplitFunc` that works with carriage returns `\r`, allowing for parsing of files created by windows software.

Closes https://github.com/asticode/go-astisub/issues/118


subtitles.go
```diff
@@ -927,3 +930,32 @@ func escapeHTML(i string) string {
 func unescapeHTML(i string) string {
 	return htmlUnescaper.Replace(i)
 }
+
+func NewScanner(i io.Reader) *bufio.Scanner {
```
asticode, line 934: Unless I'm missing something, I don't see the point of exporting this method, therefore could you rename it to `newScanner` instead?

subtitles.go
```diff
@@ -927,3 +930,32 @@ func escapeHTML(i string) string {
...
+	var scanner = bufio.NewScanner(i)
+	scanner.Split(splitLines)
+	return scanner
+}
+
+func splitLines(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
asticode, line 940: I don't really see the point of creating an extra `splitLines` function, since `newScanner` is simple enough could you add the split function directly as an anonymous function:
```go
scanner.Split(func(data []byte, atEOF bool) (advance int, token []byte, err error) {
  ...
})
```

subtitles_test.go
```diff
@@ -340,3 +340,26 @@ func TestHTMLEntity(t *testing.T) {
 		assert.Equal(t, string(c), w.String())
 	}
 }
+
+func TestNewScanner(t *testing.T) {
+	exts := []string{"vtt", "srt", "ssa"}
+	for _, ext := range exts {
+		s, err := astisub.OpenFile("./testdata/example-in-scan-line." + ext)
```
asticode, line 347: Could you rename the testdata files to `example-in-carriage-return.(srt|ssa|vtt)` instead?

asticode (changes requested):
Really nice PR!

Just a few minor changes needed 👍

subtitles.go
justin-taylor: Good point. Updated.

subtitles.go
justin-taylor: Updated.

justin-taylor:
@asticode all comments have been addressed. Let me know if there is anything else. Otherwise, can I get another tag 🙏 ?

PR #127, opened 2025-02-26
Maintain whitespace from original source

justin-taylor:
This PR solves for an issue with spacing on `LineItems`. Currently All line items are trimmed of all white space causing issues when outputting the subtitle assuming a space between each line item.

For Example: reading in
```
<i>x</i>^3
```

will output
```
<i>x</i> ^3
```
*Note the addition of a space before the `^`, which causes awkward display issues.

This PR aims to remove the need to automatically appending ` ` between each line item by maintaining white space indicated in the source. This should be safe as each line is already properly trimmed before tokenizing and parsing the html.


asticode:
That's a great PR! ❤️ 

We would need to do the same fix for `ttml`, do you feel like doing this as well?

No worries if you don't, I can also merge the PR as is 👍

justin-taylor:
@asticode, I updated the ttml handling as well

testdata/example-out-breaklines.ttml
```diff
@@ -18,15 +18,15 @@
             <p begin="00:00:02.000" end="00:00:03.000">
                 <span>Fifth line</span>
                 <br></br>
-                <span>Sixth </span>
-                <span>middle </span>
+                <span>Sixth</span>
```
asticode, line 21: We shouldn't remove the space on the right 🤔 

I remember now that there's an issue with the way we use the xml unmarshaller. Right now we're [trimming](https://github.com/asticode/go-astisub/blob/master/ttml.go#L191) all white spaces in ttml which is not correct. However I don't have a quick and easy solution for that.

I think the best would be for you to remove all the ttml-related logic you've added (sorry for that! 😱 ), I'll merge the PR (at least srt and vtt will be correct) and we'll need to find a solution for ttml after that 👍

justin-taylor: Ok, I just removed my commit with the ttml changes.

asticode:
Thanks for the PR ❤️ 

justin-taylor:
@asticode a tag would be most appreciated.

asticode:
I've created a `v0.33.0` tag 👍

srt.go
```diff
@@ -255,12 +255,8 @@ func (s Subtitles) WriteToSRT(o io.Writer) (err error) {
...
 func (l Line) srtBytes() (c []byte) {
-	for idx, li := range l.Items {
+	for _, li := range l.Items {
 		c = append(c, li.srtBytes()...)
-		// condition to avoid adding space as the last character.
-		if idx < len(l.Items)-1 {
-			c = append(c, []byte(" ")...)
-		}
 	}
 	c = append(c, bytesLineSeparator...)
```
sylwek78, line 261: -

PR #131, opened 2025-09-23
Support body div level style and region

NhanNguyen700:
Hi @asticode ,

Style and region for ttml can be specify in `body` and `div` too, please refer to testdata/example-in-style-inheritance.ttml to see the real case. Style/region will be inherited by this order `body>div>p`.

Besides, according to TTML standard, when both parent node and child node define style/region, the one in child node will win. Thus, I implement an override mechanism in your TTML parsing step to bring all the style/region from parent node down to the item (<p><\p>) node if there is no corresponding attribute in the child node. With that, the "Subtitles" object structure will not be broken and the style/region can be preserved. The side-effect is that the output TTML can be a little bit longer compare to input TTML.

### DIFF
--- a/subtitles.go
+++ b/subtitles.go
@@ -430,6 +430,109 @@ func (sa *StyleAttributes) propagateWebVTTAttributes() {
 	sa.SRTUnderline = sa.WebVTTUnderline
 }
 
+// merge - base on parent, override style attributes if defined in child
+// TODO: handle more formats than just TTML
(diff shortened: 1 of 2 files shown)


NhanNguyen700:
ping @asticode for reminding

asticode:
> @asticode , Yes, I need it, could you do me a favor and create a tag for my project?

FYI I've created a `v0.36.0` tag 👍

NhanNguyen700:
Many thanks for that!

PR #142, opened 2026-04-20
Fix panic when WebVTT cue has missing end time boundary

eric:
Previously, parsing a WebVTT line like '00:00:01.000 -->' with no end timestamp panicked with an index out of range error. Now return a descriptive error instead.


asticode:
Thanks for the PR ❤️ 

eric:
That would be great! Thanks!

asticode:
> That would be great! Thanks!

FYI I've created a `v0.40.0` tag 👍

PR #145, opened 2026-07-02
fix: nil-pointer panic in WriteToSSA when Metadata is nil

stevetoro:
### Problem
`Subtitles.WriteToSSA` panics with a nil-pointer dereference when `Subtitles.Metadata` is `nil`. `Metadata` is an optional `*Metadata` that only `ReadFromSSA` populates — `ReadFromWebVTT`, `ReadFromSRT`, `ReadFromTTML`, and `ReadFromSTL` all leave it `nil`. So reading any non-SSA format and writing SSA crashes.

Reproduced on v0.26.2. Found downstream in a caption-transcoding service where VTT→SSA conversions panicked.

```
panic: runtime error: invalid memory address or nil pointer dereference

goroutine [running]:
panic({...})
	/usr/local/go/.../src/runtime/panic.go:785 +0x132
github.com/asticode/go-astisub.Subtitles.WriteToSSA({{0xc0007c4488, 0x67, 0x8f}, 0x0, 0xc0004d1aa0, 0xc0004d1ad0}, {...})
	/root/go/pkg/mod/github.com/asticode/go-astisub@v0.26.2/ssa.go:1181 +0xba
	[... application caller frames elided ...]

```

### Root cause
In `WriteToSSA`, `newSSAScriptInfo(s.Metadata)` is already nil-safe (`if m != nil`), but the very next line dereferences `s.Metadata` unconditionally: 
```go
var v4plus = s.Metadata.SSAScriptType == "v4.00+"
```

### Fix
Nil-safe read:
```go
var v4plus = s.Metadata != nil && s.Metadata.SSAScriptType == "v4.00+"
```

### Behavior note (v4 vs v4+)
With `nil` metadata, `v4plus` is `false`, so output uses legacy `[V4 Styles]` and omits the `ScriptType:` line (the writer already guards `if len(b.scriptType) > 0`). This is valid SSA — the v4.00 flavor. I chose the minimal guard over normalizing a `v4.00+` default to keep the behavior change as small as possible for a bug fix; happy to switch to a default if you'd prefer non-SSA→SSA to emit v4.00+.

### Test
Adds `TestWriteToSSANilMetadata` (constructs `Subtitles` with a nil `Metadata` and one item, asserts `WriteToSSA` returns no error and emits `[Script Info]` + the cue text). Fails/panics on `master`, passes with the fix. Full suite still green.

### DIFF
--- a/ssa_test.go
+++ b/ssa_test.go
@@ -124,6 +124,7 @@ func TestSSAv4plus(t *testing.T) {
 }
 
 func TestWriteToSSANilMetadata(t *testing.T) {
+	// Metadata intentionally left nil
 	s := &astisub.Subtitles{
 		Items: []*astisub.Item{
 			{
@@ -133,13 +134,11 @@ func TestWriteToSSANilMetadata(t *testing.T) {
 			},
 		},
 	}
-	// Metadata intentionally left nil (as ReadFromWebVTT etc. leave it).
 
 	w := &bytes.Buffer{}
-	err := s.WriteToSSA(w)
-	assert.NoError(t, err)
-	assert.Contains(t, w.String(), "[Script Info]")
-	assert.Contains(t, w.String(), "Hello world")
+	assert.NotPanics(t, func() {
+		s.WriteToSSA(w)
+	})
 }
 
 func TestInBetweenSSAEffect(t *testing.T) {
--- a/ssa_test.go
+++ b/ssa_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io/ioutil"
 	"testing"
+	"time"
 
 	"github.com/asticode/go-astikit"
 	"github.com/asticode/go-astisub"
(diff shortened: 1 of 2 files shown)


ssa_test.go
```diff
@@ -122,6 +123,25 @@ func TestSSAv4plus(t *testing.T) {
...
+	s := &astisub.Subtitles{
+		Items: []*astisub.Item{
+			{
+				StartAt: time.Second,
+				EndAt:   3 * time.Second,
+				Lines:   []astisub.Line{{Items: []astisub.LineItem{{Text: "Hello world"}}}},
+			},
+		},
+	}
+	// Metadata intentionally left nil (as ReadFromWebVTT etc. leave it).
```
asticode: Can you move this comment before `s := ...` and make `Metadata intentionally left nil` instead ?

ssa_test.go
```diff
@@ -122,6 +123,25 @@ func TestSSAv4plus(t *testing.T) {
...
+
+	w := &bytes.Buffer{}
+	err := s.WriteToSSA(w)
+	assert.NoError(t, err)
```
asticode: Instead of the 3 below assert, can you use `assert.NotPanics(...)` instead?

asticode (changes requested):
Thanks for the PR 👍 

A few changes needed

ssa_test.go
stevetoro: Done — moved the comment above the `Subtitles` literal and simplified it to `// Metadata intentionally left nil`.

ssa_test.go
stevetoro: Done — replaced the three asserts with `assert.NotPanics(t, func() { s.WriteToSSA(w) })`.

stevetoro:
Thanks for taking the time to review this, @asticode!

asticode:
Thanks for the PR ❤️ 

stevetoro:
Yeah sure, we'll take one! We're a bit overdue for an upgrade. Thank you @asticode!

asticode:
FYI I've created a `v0.42.0` tag 👍

Issue #31, opened 2020-11-24
SRT with position properties


oreo0725:
Hi, 

I have a `.SRT` file with position properties, like following:

```
1
00:00:12,000 --> 00:00:15,123
This is the first subtitle

2
00:00:16,000 --> 00:00:18,000
Another subtitle demonstrating tags:
<b>bold</b>, <i>italic</i>, <u>underlined</u>
<font color="#ff0000">red text</font>

3
00:00:20,000 --> 00:00:22,000  X1:40 X2:600 Y1:20 Y2:50
Another subtitle demonstrating position.
```

Got this error when trying to open the file
```
astisub: line 15: parsing srt duration 00:00:22,000  X1:40 X2:600 Y1:20 Y2:50 failed: astisub: Invalid number of millisecond digits detected in 00:00:22,000  X1:40 X2:600 Y1:20 Y2:50
```

It seems that astisub not recognize such position property.
Is it possible to be supportable?

referenece:
- http://forum.doom9.org/archive/index.php/t-112947.html
- http://forum.doom9.org/archive/index.php/t-86664.html
- http://forum.doom9.org/archive/index.php/t-86664.html

asticode:
I've added limited support for this feature :

- the lib will not fail on these kind of lines anymore
- but the position won't be parsed since I haven't found an official documentation regarding this anywhere

oreo0725:
Yes, 

This seems not belong to the official SPEC and some players support such position function but some doesn't.

In my case, I use `astisub` to validate subtitle content correctness rather than getting the parsed properties.
So this fix in `v.0.10.0` is enough for me.

Thanks for quickly support.

PR #146, opened 2026-08-10
feat: handle srt tags

zoriya:
it is very common to have ssa styles embedded in srt subtitles and astisub currently doesn't handle this and leaves tags as is in the output text.

There is no official spec for this but this is widely supported (ffmpeg, vlc, mpv to name a few players that support this). Every implementation follows aegisub's wiki (http://docs.aegisub.org/latest/ASS_Tags/).
See #31 for previous discussion on the subject.

This PR adds support for the most common tags and those that astisub already handled somewhere (positions, bold...). This leaves out tags that can't be processed easily (fonts, move, custom pos and so on). We could add support for those with follow up PRs if wanted but I'm not sure if there's as much value as with those really frequent tags (like position).

### DIFF
--- a/srt.go
+++ b/srt.go
@@ -128,6 +128,13 @@ func ReadFromSRT(i io.Reader) (o *Subtitles, err error) {
 			if l := parseTextSrt(line, sa); len(l.Items) > 0 {
 				s.Lines = append(s.Lines, l)
 			}
+
+			// Alignment applies to the whole subtitle and not only to the text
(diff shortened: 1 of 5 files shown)


asticode:
Thanks a lot for the PR and sorry for the long time to review ❤️ 

zoriya:
thanks, a tag would be great so i can stop using my fork!

asticode:
FYI I've created a `v0.43.0` tags 👍

PR #147, opened 2026-08-18
fix: divide by zero panic in ReadFromSTL on unknown disk format code

ChrisJr404:
Reading an STL file whose GSI disk format code is neither `STL25.01` nor `STL30.01` panics with an integer divide by zero, so a malformed subtitle file can crash any caller of `ReadFromSTL`/`OpenFile`.

`parseGSIBlock` only set the framerate when the disk format code matched a known value and otherwise left it at `0`. When the first TTI block is then parsed, `parseDurationSTLBytes` computes `... / framerate` and divides by zero (`parseDurationSTL` on the GSI timecodes has the same problem). This fixes it by returning an error for an unknown disk format code, mirroring how an unknown character code table is already rejected in the same function.

Added a regression test (`TestSTLUnknownDiskFormatCode`) that feeds a GSI block with an unknown disk format code plus a TTI block; it panics before the change and passes after. Valid files still decode identically and the existing tests pass.

### DIFF
--- a/stl.go
+++ b/stl.go
@@ -427,6 +427,7 @@ func parseGSIBlock(b []byte) (g *gsiBlock, err error) {
 		displayStandardCode:       string(bytes.TrimSpace([]byte{b[11]})),
 		editorName:                string(bytes.TrimSpace(b[309:341])),
 		editorContactDetails:      string(bytes.TrimSpace(b[341:373])),
+		framerate:                 25,
 		languageCode:              string(bytes.TrimSpace(b[14:16])),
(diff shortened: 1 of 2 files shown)


asticode:
I think I'd rather add a default value of `25` [here](https://github.com/asticode/go-astisub/blob/master/stl.go#L428). You can still add a small test that makes sure there's no panic anymore using `assert.NotPanics()`

ChrisJr404:
Switched to a default framerate of 25 for an unknown disk format code and added the NotPanics test as suggested, thanks.

asticode:
Thanks a lot for the PR ❤️ 

PR #148, opened 2026-08-20
Preserve unparseable TTML/SRT color values instead of dropping them

jason-valenzuela:
## Problem
`newColorFromHTMLString` decoded only 6-digit `#RRGGBB` and 8 named colors. For any other value it returned an error, and both callers — the TTML reader (`ttml.go`) and the SRT reader (`srt.go`) — gate on `err == nil`, so the color was silently discarded on read and omitted on write.

That dropped values which are valid per the specs:
- `#RRGGBBAA` (8-digit, with alpha) — valid per TTML1 §8.3.2 and the form IMSC1 *mandates* for `tts:color`.
- Named colors outside the 8 recognized (e.g. `silver`, `gray`, `maroon`, `transparent`) — part of the TTML1 §8.3.2 `<namedColor>` set.
- `rgb()` / `rgba()` functional notation.

A `TTML in → TTML out` (or SRT) round-trip therefore lost styling the source explicitly set.

## Fix
- Add an unexported `Color.raw` field holding the original expression when it can't be decoded into an RGBA triple; `HTMLString()` returns it verbatim. So `#RRGGBBAA`, `rgb()/rgba()`, `transparent`, and unrecognized names now round-trip losslessly instead of being dropped.
- Decode the rest of the TTML1 named-color set (`silver`, `gray`, `maroon`, `purple`, `fuchsia`, `lime`, `olive`, `navy`, `teal`, `aqua`) into RGBA, reusing the existing `Color*` vars, alongside the 8 already handled.
- `newColorFromHTMLString` can no longer fail, so its now-dead `error` return is removed and the three call sites simplified to a direct assignment. An empty/blank expression yields a `nil` color (attribute stays omitted) rather than being flattened to `#000000`.

Decoded colors still normalize to hex on write (e.g. `white` → `#ffffff`, `silver` → `#c0c0c0`), which was already the case for the original 8 and is valid output.

## Test
`TestColorHTMLRoundTrip` covers `#RRGGBB` and named-color decoding, verbatim round-trip of `#RRGGBBAA` / `transparent` / an unrecognized name / `rgb()`, and empty/blank → nil. `gofmt`, `go build ./...`, and `go test ./...` all pass.

## Notes
Scope is intentionally minimal: `#RRGGBBAA`, `rgb()/rgba()`, and `transparent` are preserved verbatim rather than decoded into RGBA. Decoding `#RRGGBBAA` specifically would need `Color` to distinguish "alpha absent" from "alpha == 0x00" (the struct's `Alpha uint8` defaults to 0), which ripples into `SSAString`/`WebVTTString`; that's a larger change best left as a follow-up.


jason-valenzuela:
Bumping this PR @asticode, let me know if there's any additional context you'd like for me to provide!

subtitles.go
```diff
@@ -144,6 +144,13 @@ func (i Item) String() string {
 // Color represents a color
 type Color struct {
 	Alpha, Blue, Green, Red uint8
+
+	// raw preserves the original TTML/SRT color expression when it cannot be
+	// decoded into an RGBA triple (e.g. #RRGGBBAA with alpha, rgb()/rgba()
+	// functional notation, or a named color outside the recognized set).
+	// HTMLString returns it verbatim so these values round-trip losslessly
+	// instead of being silently dropped.
+	raw string
```
asticode: I'd rather rename this to `rawHTML` for now since we may need apply your fix to both `SSA` and `WebVTT` colors as well in the future.

asticode (changes requested):
Thanks for the PR ❤️ 

Just one minor change needed 👍

subtitles.go
jason-valenzuela: Agreed! Just pushed up that change.

asticode:
Thanks for the PR ❤️ 

jason-valenzuela:
Excellent, thank you! Yes, I'd like a tag.

asticode:
> Excellent, thank you! Yes, I'd like a tag.

FYI I've created a `v0.44.0` tag 👍

PR #150, opened 2026-09-14
fix: don't treat block-header lines as headers inside cue text

jason-valenzuela:
## Summary
`ReadFromWebVTT` misparses header-only lines when they appear inside a cue's text, silently corrupting the caption. The top-level line `switch` starts a NOTE / REGION / `Region:` / STYLE / X-TIMESTAMP-MAP block whenever the line matches, without checking whether we're currently accumulating cue text. So a cue whose body is one of those keywords loses data.

## Problem
Given a cue whose text is exactly `REGION` (all-caps text is common in broadcast captions):

```
WEBVTT

00:01:00.000 --> 00:02:00.000
REGION
```

`REGION` is consumed as the start of a new region block and the cue body is dropped. Related shapes are worse:

- `REGION` followed by a `key: value` line injects a bogus region into the header.
- A cue body line `Region: id=foo` is promoted to a real header region (and a malformed one aborts the whole parse with `Invalid region style`).
- A cue body line `STYLE...` / `NOTE ...` is swallowed as a style/comment block.
- A cue body line `X-TIMESTAMP-MAP=...` overwrites `Metadata.WebVTTTimestampMap` and drops the cue text.

Per the WebVTT spec, once a cue's payload begins, these keywords are literal cue text, not block introducers.

## Fix
Compute `inCue := blockName == webvttBlockNameText` once per line and add `&& !inCue` to the five block-start cases. Inside a cue, those lines now fall through to the default branch → `parseTextWebVTT`, i.e. they're kept as literal cue text. Header and between-cue behavior is unchanged (`blockName` is `""` there, so the guard is a no-op). Leading/trailing whitespace variants are already handled by the existing up-front `strings.TrimSpace`.


jason-valenzuela:
@asticode Found this issue when I was upgrading to the latest release, let me know what you think!

asticode:
Thanks for the PR ❤️ 

jason-valenzuela:
@asticode You rock, thanks! Yes, we will need a tag.

asticode:
> @asticode You rock, thanks! Yes, we will need a tag.

FYI I've created a `v0.45.0` tag 👍

